PR #10, opened 2015-08-10
Fix test

zorkow:
Fixes the default value for test equations.

### DIFF
--- a/Semantics-Lab/Semantic-Lab.js
+++ b/Semantics-Lab/Semantic-Lab.js
@@ -132,10 +132,6 @@ var Lab = {
     this.Typeset();
   },
   //
-  // Current test equation.
-  //
-  Current: 0,
-  //
   // Select next test equation.
   //
   Next: function() {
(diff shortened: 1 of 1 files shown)


pkra:
LGTM.

dpvc:
This change is not correct.  The Current value was set earlier in [line 192](https://github.com/mathjax/MathJax-RespEq/blob/6f1ddd73a265f7f04b6c03370ca3fbe9fb742a2d/Semantics-Lab/Semantic-Lab.js#L192), to the value that was passed to it from the URL options (the last saved value).  Setting it to 0 will override that.

Can you say what problem you are trying to solve?  Some browsers do their own form input caching, and it may be that that is causing some issues.  The correct solution may be to set the value of the input field explicitly to the `lab.Current` value, not set `lab.Current` to 0.

@pkra, you were a bit quick off the draw to merge this.  Did you test it out to see if it worked as expected?  I didn't have a chance to look through the code to see what the changes were, even.

zorkow:
It gave me a NaN as value return value for this.Current.
Even the initial value of Lab.Current was NaN.

Volker

dpvc:
What does the URL look like?

dpvc:
Perhaps changing line 192 to be

      Lab.example.value = defaults[1]; Lab.Current = parseInt(defaults[1]) || 0;

would do the trick.

zorkow:
Right, I can now see the assignment (that statement is double parked ;-)

Still parseInt("") is NaN.

Maybe change line 172 to
  var defaults = [null,0,

dpvc:
Sure, that works, too.

PR #11, opened 2015-08-10
Corrects setting of test example default.

zorkow:


dpvc:
I hate to say it, but there is still a problem; not with the code, but with the branching.  This is a branch from `gh-pages`, but it should be from `master` and to `master` (since we don't have a `develop` branch here).  Then `master` should be merged into `gh-pages`.  I'd like to try to maintain the `master` (or later the `develop` branch if we need it) to be the place where changes to the code goes, and then other branches (like the `gh-pages` or other special-purpose branches) are targets for pulls from `master`, but not sources of pulls *to* `master`.

Can you rebase the change onto a branch originating from `master`?  Thanks.

zorkow:
Sure, no problem. I had originally branched off of master (actually push
directly by accident...), but Peter alerted me that the actual webpages I
wanted to fix are generated from the gh-pages branch.
So the current version already helped with that.

Best,
Volker

PR #20, opened 2015-10-16
Fix lab problems

dpvc:
This patch fixes the timing issue with SRE in the Sementic-Lab.js file (at least I am not able to get it to fail any longer).  It also fixes the initialization problem with the Exampes drop-down menu in the TeX lab, and removes the InputHooks.js file since that has been merged into the develop branch.


dpvc:
I'm still working on the HTML-CSS dependency in Semantic-Collapse.js.  It turns that there are a number of places that are dependent on the HTML-CSS renderer, and I haven't worked out all the replacements yet.

zorkow:
Worked for me in all modes.

zorkow:
Finished according to today's call.

PR #21, opened 2015-10-17
Add support for collapsing all output modes

dpvc:
This makes the collapsing work in all output modes (other than PreviewHTML).

It also fixes the problem where NativeMML output is collapsed initially when it shouldn't be.

### DIFF
--- a/Semantics-Lab/Semantic-Collapse.js
+++ b/Semantics-Lab/Semantic-Collapse.js
@@ -210,7 +210,7 @@
     //  the equation to be expanded or collapsed further.
     //
     collapseActions: function (SRE,state) {
-      var w = SRE.width ,m = w, M = 1000000;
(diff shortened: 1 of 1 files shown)


zorkow:
Observations:
SVG: Collapse works stepwise when using the slider. Interactively, when I try to uncollapse sub-expressions or collapse parts of the formula, I always get a collapse of the full formula. Maybe there's a problem with bubbling, or I just can't hit the right rectangle. (FF + Chrome + Opera on Linux)

NativeMML: Top level is now uncollapsed. Collapse works stepwise when using the slider. Interactively, it does the same to-and-fro-ing between top-level and un/collapsed subexpressions. (FF on Linux)

zorkow:
Update: 
SVG: The Collapse problem was because I was running with a local MathJax and not the rawgit!

PR #26, opened 2015-10-29
Shrink enriched MathML display

pkra:
This keeps making me mad...

### DIFF
--- a/Semantics-Lab/Semantics-Lab-TeX.html
+++ b/Semantics-Lab/Semantics-Lab-TeX.html
@@ -165,7 +165,7 @@ Enriched Math:
 <span id="range_output">100%</span>
 <br/></br>
 Enriched MathML:
-<div style="border:1px solid; padding:3px">
+<div style="border:1px solid; padding:3px; height:20px; overflow:auto;">
 <pre id="mathml"></pre>
 </div>
 </p>


dpvc:
What keeps making you mad?  The fact that the enriched MathML is shown?  Perhaps we should add a checkbox to control whether it is displayed or not, but reducing its size to one line seems inappropriate to me.  This display was critical for me when I was working on the collapsing, as I needed to be able to inspect the complexity values, as well as the structure created by the enrichment process.  I assume that it is useful to Volker as well in his testing.  It would be incredibly difficult to work with if you can only see one line at a time and had to scroll it by hand to find anything.

I'm not sure what problem you are trying to solve with this pull request.  The lab was designed as a debugging tool for Volker and me.  If you have another use that you want to put it to, perhaps we need another version of it for that.

pkra:
> What keeps making you mad?

Life, the universe and Everything. Oh wait.

Sorry for framing it very poorly. When trying to debug the walker, the enriched MathML takes up a lot of screen real estate, especially on a small VM screen, with an inspector open.

> reducing its size to one line seems inappropriate to me.

You're right.

dpvc:
PS, I'm fine with the `overflow:auto`, it's the `height:20px` that I object to.

Also, both the TeX and MathML labs should get the same change.

dpvc:
I think an option to turn of showing the enrichment might be a better solution.

Issue #32, opened 2015-11-18
SVG maction class name


zorkow:
@dpvc 
I would have expected the collapsible elements in SVG to have class mjx-svg-maction.
However, they have no classname at all, so hover and flame do not work.

dpvc:
Remember that by default the SVG output doesn't preserve the MathML structure (it collapses groups to reduce the complexity of the output).  So it may remove the SVG elements that correspond to the individual MathML elements.  If you want to have the structure preserved in SVG output, you need to set `addMMLclasses` to true in the SVG configuration block.  Then you will have the elements marked by class names like `mjx-svg-maction`, and the group structure will reflect the MathML tree.

pkra:
This can be closed?

dpvc:
I think so.  Pull 33 should have resolved the issue.

PR #33, opened 2015-11-19
Adds MML classes in SVG configuration.

zorkow:
Addresses issue #32.

### DIFF
--- a/Semantics-Lab/Semantic-Collapse.js
+++ b/Semantics-Lab/Semantic-Collapse.js
@@ -747,6 +747,7 @@ MathJax.Hub.Register.StartupHook("HTML-CSS Jax Ready",function () {
 //  Specific implementations for SVG output
 //
 MathJax.Hub.Register.StartupHook("SVG Jax Ready",function () {
+  MathJax.Hub.Config({SVG: {addMMLclasses: true}});
   MathJax.OutputJax.SVG.Augment({
     getMetrics: function (jax) {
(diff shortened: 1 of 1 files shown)


pkra:
LGTM.

dpvc:
I was going to say that an alternative to this would be to have the collapsing code (which inserts the mactions) mark the mactions with its own class name, so that (a) you don't need to turn on MML classes (so that the SVG is reduced in complexity and size), and (b) you can looks for the same class name in all the output modes, since it will be kept the same for them all.

pkra:
@dpvc happy to revert and follow your suggestion.

zorkow:
I understand the reasoning. However, the additional structure makes it easier for the walker. In general for SVG accessibility more structure is better than less. 
What is the additional complexity? Is it a constant factor?

dpvc:
OK, sounds good.  Leave it as is.

dpvc:
> What is the additional complexity? Is it a constant factor?

Sorry forgot to answer this.  No, I don't think it is constant, but haven't tried to measure it exactly.  It consists of removing nested groups that are not necessary. so an mrow consisting of an mn, an mo, and an mn with `addMMLclasses` set to `true` would normally have a `g` element for each of the MathML elements, but without `addMMLclasses`, only the outer `g` for the `mrow` would be retained, and it would contain three path elements (assuming one character for each of the token elements).  If that `g` were a child of some other `<g>` higher up in the MathML structure, it could also be removed.

PR #36, opened 2015-11-27
Adds timeout for initial element.

zorkow:
That could take care of the initial message problem. It was indeed a timing issue. 
Tested with Win7/IE10/NVDA. 
@pkra could you please have a look at other combinations?

### DIFF
--- a/Semantics-Lab/Assistive-Explore.js
+++ b/Semantics-Lab/Assistive-Explore.js
@@ -15,6 +15,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
     flamer: null,
     speechDiv: null,
     enriched: {},
+    focusEvent: MathJax.Hub.Browser.isFirefox ? 'blur' : 'focusout',
     //
     // Configurations.
     //
@@ -65,7 +66,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
         if (math) {
           math.onkeydown = Explorer.Keydown;
           math.addEventListener(
-            MathJax.Hub.Browser.name === 'Firefox' ? 'blur' : 'focusout',
+            Explorer.focusEvent,
             function(event) {
               if (Explorer.walker) Explorer.DeactivateWalker();
             });
@@ -182,8 +183,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
       Explorer.walker = new constructor(math, speechGenerator);
       Explorer.GetHighlighter(.2);
       Explorer.walker.activate();
-      setTimeout(function() {Explorer.Speak(Explorer.walker.speech());}
-                 , 100);
+      Explorer.Speak(Explorer.walker.speech());
       Explorer.Highlight();
     },
     //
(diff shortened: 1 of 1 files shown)


pkra:
Will do.

pkra:
I can confirm that NVDA now catches the first speech string on Win7 FF, Chrome. JAWS does not.

But more seriously, I cannot walk at all using that branch. Do I need to merge other branches into this?

pkra:
FWIW, merging the other PRs locally did not help with the broken walker (though I get earcons so the walker must think it's at an edge)

pkra:
I just realized that JAWS did not have the initial-speech-string problem in the past :-(

On Wind 10, JAWS only voices the initial speech string on Chrome.

pkra:
Hm... I think the support matrix is wrong and JAWS did not voice the initial speech in FF on Win 10. At least I can't get it to with the current main branch.

pkra:
Were there changes to SRE?

zorkow:
Walker should work. It works for me at least.

zorkow:
Apart from the bug fix for the embellished fences, there were no changes to SRE.

pkra:
The walker does not work for me on any browser or OS anymore (on https://mathjax.github.io/MathJax-RespEq/Semantics-Lab/Semantics-Lab-TeX.html)

pkra:
FWIW, with or without screenreaders it does not work.

The same applies to the `timeout_first_message` branch, with or without the other branches merged in.

zorkow:
It works perfectly fine for me. On all branches. Sorry I am at a loss, as I am not sure where I could start the debugging.
@dpvc do you have any problems?

zorkow:
Just tried Chrome and Safari on the missus's MacBook. Both worked fine.

pkra:
D'oh! Of the highest order. I was stuck on the dummy walker :-(

Please excuse my total failure. Will test properly now...

pkra:
I did some initial tests on Win8.1 using IE11, FF, Chrome.

This solves the initial speech string bug in NVDA and works with JAWS as well.

FWIW, the NVDA "sometimes lose first item in level" bug still persists, as does the JAWS "I start with lots of blanks until I magically stop doing that".

The earcon works with NVDA on FF and Chrome but I couldn't get it to work with JAWS on either (and the ogg won't work anywhere else). Will test more tomorrow.

zorkow:
We can certainly change the earcon to something else. mp3 is probably a good alternative, but we may need to do some browser sniffing re FF and Opera (https://en.wikipedia.org/wiki/HTML5_Audio).

zorkow:
This now has a selective timeout for Chrome & PC.

pkra:
Will test this now. FYI the current PRs are diverging.

pkra:
Preliminary summary (without WindowsXP, see above): this branch fixes NVDA's Chrome issues but no other issue. 

On the bright side, I didn't spot any regressions.

pkra:
In case it wasn't clear, I think it's ready to merge.

PR #37, opened 2015-11-27
Adds signal for explorer ready.

zorkow:
Additional signal when the explorer is fully loaded. For anything building on top of that module.


pkra:
LGTM.

PR #38, opened 2015-11-27
Adds a simple earcon to indicated boundaries of expressions.

zorkow:
Adds a very simple earcon. See https://github.com/zorkow/speech-rule-engine/issues/30speech-rule-engine


pkra:
Awesome, will test. (But ogg only supported in Chrome and Firefox.)

zorkow:
This should now also support non-ogg browsers. PTAL.

pkra:
I did primarily a browser test (since AT isn't really needed to check this but I did a few combinations as well, one for each AT). Works as expected in all browsers we support.

PR #42, opened 2015-12-02
add a11y demo

pkra:
This adds the demo that's currently only on the gh-pages branch.

It uses a hacked copy of Semantic-Collapse.js so that there's no collapse but the walker works. Happy to refactor once we can split things up more cleanly.

### DIFF
--- a/Semantics-Lab/walker/Semantic-Collapse-walker.js
+++ b/Semantics-Lab/walker/Semantic-Collapse-walker.js
@@ -1,0 +1,1042 @@
+//
+//  A filter to convert the enhanced MathML to MathJax internal format
+//  so we can display it, while adding maction elements for parts that
+//  can be collapsed.  We determine this based on a "complexity" value
+//  and collapse those terms that exceed a given complexity.
+//
+//  The parameters controlling the complexity measure still need work.
+//
+(function () {
+  var MML;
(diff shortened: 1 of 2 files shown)


Semantics-Lab/walker/Semantic-Collapse-walker.js
```diff
@@ -0,0 +1,1042 @@
...
+    //
+    //  If there is a specific routine for the type, do that, otherwise
+    //  check if there is a complexity cut-off and marker for this type.
+    //  If so, check if the complexity exceeds the cut off, and
+    //    collapse using the appropriate marker for the type
+    //  Return the (possibly modified) MathML
+    //
+    Collapse: function (node,mml) {
+      var type = mml.attr["data-semantic-type"];
+      if (false) {
```
zorkow: This should be guarded by a global flag.

zorkow:
Unless we really need this as a temporal snapshot, I feel we should not have an internal fork of the semantic-collapse file.
We should introduce a global flag and rewrite the collapse function to contain code of the nature:
if (collapseFlag && type) {

pkra:
At the time, I didn't have time to properly understand the code, so I just hacked this together. Happy to get some input.

PR #45, opened 2015-12-09
New a11y demo

zorkow:
This is the a11y demo using a flag to switch off collapse functionality. It might still have some timing issues, although I never came across any.
@pkra can you have a look if that works for you?


pkra:
Thanks for fixing this. I do see timinig issues caused by setting up the options in https://github.com/mathjax/MathJax-RespEq/blob/new_a11y_demo/Semantics-Lab/Semantic-Lab.js#L236-L239. Will try to narrow it down.

pkra:
I don't actually understand how https://github.com/mathjax/MathJax-RespEq/blob/new_a11y_demo/Semantics-Lab/Semantic-Lab.js#L144-L146 is supposed to work but replacing it with `MathJax.Hub.Queue(MathJax.Extension.Explorer.setExplorerOption(key, value));` seems to solve the issue. I'm guessing `Semantic-Lab` should wait for `Assistive-Explorer` to be safe but it's not a proper extension anyway right now, correct?

zorkow:
#47 Should take care of problems with setting options.

zorkow:
Damn, forgot to rebase before pushing. It now includes #47

zorkow:
Decided to merge all pending PR into this for ease of testing.

Semantics-Lab/walker/index.html
```diff
@@ -0,0 +1,195 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>MathJax Accessibility Tool</title>
+  <script type="text/x-mathjax-config">
+    MathJax.Hub.Config({ menuSettings: {CHTMLpreview: false}, MMLorHTML: {MSIE: "HTML"},  "AssistiveMML.js": {disabled: true} });
```
pkra: This one

pkra:
LGTM

zorkow:
Unless there are any objections, I'll merge.

PR #47, opened 2015-12-09
Fixes timing issue by queuing explorer options until ready signal.

zorkow:
This should now take care of the timing issue in a slightly more elegant way than guarding every call to setexploreroptions.


pkra:
> a slightly more elegant way

For a random value of elegant ;-)

LGTM.

PR #48, opened 2015-12-09
Add aria label

zorkow:
Adds an aria role application and aria label to announce navigatable math.


Semantics-Lab/Assistive-Explore.js
```diff
@@ -56,6 +56,14 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
       }
     },
     //
+    // Adds Aria attributes.
+    //
+    AddAria: function(math) {
+      math.setAttribute('role', 'application');
+      math.setAttribute('aria-label',
+                        'Navigatable Math. Explore with shift space.');
```
dpvc: I think the correct word is "Navigable".  "Navigatable" doesn't pas spell checking.

zorkow: For what it's worth: http://www.definitions.net/definition/navigatable

zorkow: It's a bit like clickable which usually does also not spell check.

pkra:
Quick test with NVDA works well, No more mode switching necessary, label is read on FF and IE11 Win10. Also first-string is read on IE11 and Chrome (though chrome does not read label and needs mode switching)

pkra:
Virtually identical results in JAWS on Win10 (modulo their usual differences)

pkra:
Addendum: JAWS+Chrome seems to read the window title instead of the label.

WindowEyes on Win10. FF works well but instead of the label, it voices "appllication application main window". But Chrome has deteriorated; speech seems to get cut-off by something else.

pkra:
Windows Eyes still works flawlessly in Edge and IE11 on Win 10.

I do think the label is too long since it will be voiced on every equation.

pkra:
Win7 support has deteriorated badly. NVDA and Window Eyes don't voice anything on Chrome and FF anymore.

pkra:
Hm... Might be a broken Win7 VM. Will try a fresh one later.

pkra:
Ok, NVDA on Win7. FF, Chrome behave almost as on Win10. Except that first speech string is lost on FF as well (on the bright side: the second Chrome bug with first elements on levels is fixed). IE9 voices labels, misses initial speech text.

pkra:
Wait... how did I get it to run in IE9? I must have been dreaming that one.

Anyway, JAWS works as in Win10 for FF and Chrome (no IE9).

zorkow:
> I do think the label is too long since it will be voiced on every equation.

So what could we announce? We should mention the key combination.

pkra:
I don't think we should mention the key combo if we voice this at every equation without the user being able to modify this behavior. It's as annoying as alerts if you're reading a whole text.

zorkow:
Just "navigatable math" then?

A total alternative would be to announce the entire Math expression ending with navigatable math.

pkra:
> A total alternative would be to announce the entire Math expression ending with navigatable math.

Right, that came to mind given how some AT reads math. Personally, I never liked that much especially since the important information comes at the very end. I would suggest something very short possible like "math" when it comes to every equation. 

I think we'll need some feedback from users on this.

As I wrote elsewhere, we could also think about announcing some information in the aria live region when the tool starts up.

pkra:
Let's merge this and take it from there.

zorkow:
Shortened the Aria label. I will refactor the entire live region shortly and then make an initial announcement, if in page.

PR #49, opened 2015-12-10
Changes live region

zorkow:
Adds a single live region permanently asap.
Gets rid of some of the timing problems with NVDA (e.g., initial speech string not read on Win8.1 IE11).
Does not help re Chrome problem of sub formulas not read.


pkra:
LGTM

PR #52, opened 2015-12-16
Refactor aria region

zorkow:
Refactors ARIA region and adds an initial announcement about navigatable Maths expressions.

### DIFF
--- a/Semantics-Lab/Assistive-Explore.js
+++ b/Semantics-Lab/Assistive-Explore.js
@@ -79,7 +79,8 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
     flamer: null,
     speechDiv: null,
     enriched: {},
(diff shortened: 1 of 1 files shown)


zorkow:
I've tested it with ChromeVox, Jaws and NVDA (Win 7), Chrome, IE, Firefox. Results similar to matrix. ChromeVox has the occassional D I V problem.

pkra:
https://github.com/mathjax/MathJax-RespEq/wiki/Support-Matrix-a11y-tool#pull-54 

Only issue is the regression in WindowEyes with Chrome.

I think we should go with this; the improvements far outweigh this regression.

pkra:
Thanks. WinXP, Earcon over http in Chrome is fixed now.

zorkow:
So that's ready to merge then.

PR #56, opened 2015-12-17
Adds a timeout for the initial announcement.

zorkow:
This now seems to work with most SR/BR combinations. 

In Jaws/Chrome one needs  to switch manually to virtual cursor for navigation.


Semantics-Lab/Assistive-Explore.js
```diff
@@ -12,7 +12,8 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
   var LiveRegion = MathJax.Extension.LiveRegion = MathJax.Object.Subclass({
     div: null,
     Init: function() {
-      this.div = LiveRegion.Create('assertive');
+      this.div = LiveRegion.Create('assertive',
+                                   {fontSize: '10px', color: '#000000'});
     },
```
zorkow: @pkra This is the bit we have talked about earlier. 
Eventually this should become {fontSize: '1px', color: '#FFFFFF'}.

dpvc: I'm not sure that is going to be sufficient, if your goal is to make it not visible to the sighted, but still available to AT.  Not everyone uses white backgrounds, so they may see a line of white does or something.  I might recommend using something like

```
position:absolute; top:0;
height: 1px; width:1px;
padding: 1px;
overflow: hidden;
```

so that 

1.  The element is taken out of the page flow.
2.  Its contents will not affect the width or height of the page.
3.  Its contents will not be visible to the sighted.

These could be set in a style for `MathJax_SpeechOutput`, for example.

pkra:
For the record. I don't think this is necessary for NVDA but I still want to go back and check if Window Eyes's problem on Chrome might improve.

pkra:
Ok, I checked. This does not help WindowEyes. Let's reject.

PR #59, opened 2016-01-06
Patch struik

pkra:


pkra:
This seems to be ok. Would love to merge it into gh-pages afterwards for tomorrow's talk.

Issue #58, opened 2016-01-04
struik demo broken


pkra:
https://mathjax.github.io/MathJax-RespEq/Semantics-Lab/Struik.html gives me processing errors everywhere.

Maybe 2.6 related breakage?

dpvc:
Looks like the Struik page is badly out of date.  It still loads the InputHooks file that was removed some time ago, and (more importantly), calls on the old semantics.js rather than sre_mathJax.js.

I don't remember what functionality we wanted for that page.  Do you want the full walker, or just the responsive collapsing?

dpvc:
PS, the error message causing the problem is

    'undefined' is not an object (evaluating 'sre.Enrich.semanticMathmlSync')

so it is related to the semantic enrichment, and may be due to loading the wrong semantic enrichment file.

dpvc:
OK, I've added a `patch-struik` branch that fixes the call to `sre` that is causing the processing error.  This makes the page display, but it looks like the adaptive stuff isn't working.  I haven't looked into that yet.

zorkow:
I've fixed the patch to point to the correct library now (it's no longer semantics but enrich, the semantics library was still a left over.) 
There's still a timing issue, I could not fix or understand for that matter. Here's an attempt at an explanation:

If the initial rendering is finished before the MML.mbase.Augment functions are set, the enrichment does not have any effect.
Even if I call a rerender manually it is not used. So the jax.enriched actually contains the correctly enriched MathML expression, it is not used.

Currently this does not happen, since the enrich.js immediately sends a Sre Ready message, since speech is switched off and it assumes it is never used. But that's not ideal.

dpvc:
> There's still a timing issue, I could not fix or understand for that matter.

I've looked into it and think I know what is going on.  (More below)

> If the initial rendering is finished before the `MML.mbase.Augment` functions are set, the enrichment does not have any effect.  Even if I call a `rerender` manually it is not used.

You are correct about the timing of the `MML.mbase.Augment`.  The reason that `Rerender()` doesn't help is that the enrichment is done during the input phase, and `Rerender()` only redoes the output phase, not the input phase.  You would need to do `Reprocess()` to get both phases done.

[...]

PR #61, opened 2016-01-24
Make scripts into extensions

dpvc:
This pull request turns the scripts like `Semantic-MathML.js` into actual extensions, so that they can be loaded by MathJax using its synchronization process for extensions.  This should resolve the remaining timing issue from #58, without requiring any chances to SRE.  The `Semantic-MathML` extension now forces the loading of SRE, and the `Semantic-Collapse` forces loading of `Semantic-MathML` (so that users don't have to know the internal dependancies).  I've also added a custom path (`[RespEq]`) so that the extensions can be loaded via `[RespEq]/Semantic-Math`, and so on.  Similarly, I added `[SRE]` for where the SRE is loaded by `Semantic-MathML`.  If it is not specified by the page, then the current progressiveaccess.com address is used, but this allows it to be loaded locally or from the third-party repository if we want.

I've also reorganized the structure of the Semantic-Lab directory to separate out the extensions from the examples from the labs themselves, and renamed the labs to make their names less redundant.  I fixed up the `*-linebreak` labs to work with the current versions of the extensions (though I'm not sure why they seem to be deleted and re-added below, as I renamed them just like I did the others). 

The functionality of the `Semantic-Action-*` scripts have previously been moved to the SRE, so I've removed those files.

Let me know if you have any problems with the new organization.  Otherwise, I'll merge the changes in a day or two.


zorkow:
There is now a problem with the walker demo that it does not switch of collapse.

dpvc:
OK, give it a try now.  It should be disabled now.

zorkow:
LGTM.

zorkow:
I hit the wrong button!

PR #72, opened 2016-02-10
Removes relative path for tex lab.

zorkow:


dpvc:
Sorry, that was left over from my testing.

zorkow:
NP. The real problem was that it worked well with my local paths as well!

Issue #57, opened 2015-12-18
[a11y demo] make textarea accessible


pkra:
Via stevefaulkner on a11y slackers. the textarea needs associated label `<label for="input">Enter TeX code here:</label>`

Sigh.

zorkow:
Done.

PR #75, opened 2016-02-10
Two versions of the labelled text area.

zorkow:
This is a step towards fixing issue #57.

In TeX.html, we add an aria-label. This way the textarea is announced and the original description is also reachable from the screen reader.

In MathML.html, we turn the description into a label. This takes it out of the normal flow for the screenreader it will only be read twice if the entire table is selected.

Personally I think the latter is good, but have a look.


zorkow:
Unless there any objections I'd go with the version implemented in MathML.html.

dpvc:
I'm OK with either one.  I haven't been doing the testing with the screen readers, so don't know which is more effective or more widely usable.

Issue #67, opened 2016-02-01
[a11y walker] improving UX for documents / AT reading modes


pkra:
As an AT user, I want to read documents with mathematics the same way I read other documents, using my AT's smart modes, e.g., to read the entire document or read it line-by-line.

This experience falls short because, e.g., NVDA will (sensibly) read the aria-label ("math") when reading line-by-line, i.e., "When math holds, we get math which implies math". 

While some navigation is availabe to reach back (ctrl+tab to go to previous focusable), it is not always reliable (jumping too far behind). Also, once you're on the focused element, I couldn't figure out if there's a way to get back into the flow.

Some ideas.

* talk to AT community on expected / best-of-class behavior.
* add complete speech text as aria-label (preceded by "math", perhaps, just like headings are preceded by "heading level n")
* consider if we could/should/would get AT to stop at each equation (treating them as newlines, even inline math), and enable smooth UX for activating and continuing(/skipping) the walker.

What else could we do?

zorkow:
The into-struik branch contains a version where the math speech is pushed into the aria-label.
At least ChromeVox stops after a certain number of characters.
Maybe we could put the shortened text in there.

zorkow:
This is  addressed in PR #92 
Labels contain summaries.

PR #76, opened 2016-02-12
Accessibility menu

zorkow:
* Integrates the AT options into a submenu.
* Revamps the subtitling and adds an option to switch it on or off.
* There is probably a nicer way to deal with config options and make the available to the menu, but I don't really know how.
* It does not yet integrate well with the Lab and combobox options. I could duplicate code similar to renderer setting, but would like an explanation of the following code snippets first from @dpvc 

  //
  //  The renderer selection
  //
  setRenderer: function (mode,skipUpdate) {
    MathJax.Hub.Queue(["setRenderer",MathJax.Hub,mode,"jax/mml"]);
    if (!skipUpdate) MathJax.Hub.Queue(["Rerender",MathJax.Hub]);
  },
  

//
//  Hook into menu renderer changes
//
MathJax.Hub.Register.StartupHook("MathMenu Ready",function () {
  MathJax.Extension.MathMenu.signal.Interest(function (message) {
    if (message[0] === "radio button") {
      var renderer = message[1].value;
      if (String(renderer).match(/^(HTML-CSS|CommonHTML|PreviewHTML|NativeMML|SVG)$/)) {
        Lab.renderer.value = renderer;
      }
    }
  });
});

### DIFF
--- a/Semantics-Lab/Lab.js
+++ b/Semantics-Lab/Lab.js
@@ -107,7 +107,6 @@ var Lab = {
     if (!skipUpdate) {
       MathJax.Hub.Queue(
         ["Reprocess",this.jax[1]],
-        ["ShowMathML",this],
         ["CollapseWideMath",MathJax.Extension.SemanticCollapse]
       );
     }
@@ -207,6 +206,11 @@ MathJax.Hub.Register.StartupHook("Explorer Ready",
                                    Lab.ASSISTIVE = MathJax.Extension.Assistive;
                                    Lab.executeExplorerOptions();
                                  });
+
+//
+// Hook into reprocess as this can be also triggered from the menu.
+//
+MathJax.Hub.Register.MessageHook('End Reprocess', ['ShowMathML', Lab]);
 
 //
 //  Hook into menu renderer changes
--- a/Semantics-Lab/Lab.js
+++ b/Semantics-Lab/Lab.js
@@ -14,8 +14,8 @@ var Lab = {
     background: "blue",
     foreground: "black"
   },
-  // The explorer extension
-  EXPLORER: null,
+  // The assistive extension
+  ASSISTIVE: null,
   //
   //  The TeX code for the examples menu
   //
@@ -48,10 +48,10 @@ var Lab = {
       String(window.location).replace(/\?.*/,"")+"?"
         +[this.example.value, this.width.value, this.renderer.value,
           this.collapse, this.overflow,
-          Lab.EXPLORER.config.walker,
-          Lab.EXPLORER.config.highlight,
-          Lab.EXPLORER.config.background,
-          Lab.EXPLORER.config.foreground,
+          Lab.ASSISTIVE.config.walker,
+          Lab.ASSISTIVE.config.highlight,
+          Lab.ASSISTIVE.config.background,
+          Lab.ASSISTIVE.config.foreground,
           ""].join(';')
         +escape(this.input.value);
   },
@@ -144,9 +144,9 @@ var Lab = {
   //
   explorerOptions: [],
   executeExplorerOptions: function() {
-    while (Lab.EXPLORER && Lab.explorerOptions.length > 0) {
-      Lab.EXPLORER.setExplorerOption.apply(Lab.EXPLORER,
-                                           Lab.explorerOptions.pop());
+    while (Lab.ASSISTIVE && Lab.explorerOptions.length > 0) {
+      Lab.ASSISTIVE.setOption.apply(Lab.ASSISTIVE,
+                                    Lab.explorerOptions.pop());
     }
   },
   setExplorerOption: function(key, value) {
@@ -204,7 +204,7 @@ MathJax.Hub.Register.MessageHook("New Math",["NewMath",Lab]);
 //
 MathJax.Hub.Register.StartupHook("Explorer Ready",
                                  function() {
-                                   Lab.EXPLORER = MathJax.Extension.Explorer;
+                                   Lab.ASSISTIVE = MathJax.Extension.Assistive;
                                    Lab.executeExplorerOptions();
                                  });
 
--- a/Semantics-Lab/Lab.js
+++ b/Semantics-Lab/Lab.js
@@ -216,14 +216,14 @@ MathJax.Hub.Register.StartupHook("MathMenu Ready",function () {
     if (message[0] === "radio button") {
       var variable = message[1].variable;
       if (variable === 'renderer') {
-        Lab.renderer.value = value;
+        Lab.renderer.value = message[1].value;
         return;
       }
       if (String(variable).match(/^Assistive-/)) {
         var key = String(variable).replace('Assistive-', '');
         var element = document.getElementById(key);
         if (element) {
-          element.value = value;
+          element.value = message[1].value;
         }
       }
     }
--- a/Semantics-Lab/Lab.js
+++ b/Semantics-Lab/Lab.js
@@ -214,12 +214,11 @@ MathJax.Hub.Register.StartupHook("Explorer Ready",
 MathJax.Hub.Register.StartupHook("MathMenu Ready",function () {
   MathJax.Extension.MathMenu.signal.Interest(function (message) {
     if (message[0] === "radio button") {
-      var value = message[1].value;
-      if (String(value).match(/^(HTML-CSS|CommonHTML|PreviewHTML|NativeMML|SVG)$/)) {
+      var variable = message[1].variable;
+      if (variable === 'renderer') {
         Lab.renderer.value = value;
         return;
       }
-      var variable = message[1].variable;
       if (String(variable).match(/^Assistive-/)) {
         var key = String(variable).replace('Assistive-', '');
         var element = document.getElementById(key);
--- a/Semantics-Lab/Lab.js
+++ b/Semantics-Lab/Lab.js
@@ -214,9 +214,18 @@ MathJax.Hub.Register.StartupHook("Explorer Ready",
 MathJax.Hub.Register.StartupHook("MathMenu Ready",function () {
   MathJax.Extension.MathMenu.signal.Interest(function (message) {
     if (message[0] === "radio button") {
-      var renderer = message[1].value;
-      if (String(renderer).match(/^(HTML-CSS|CommonHTML|PreviewHTML|NativeMML|SVG)$/)) {
-        Lab.renderer.value = renderer;
+      var value = message[1].value;
+      if (String(value).match(/^(HTML-CSS|CommonHTML|PreviewHTML|NativeMML|SVG)$/)) {
+        Lab.renderer.value = value;
+        return;
+      }
+      var variable = message[1].variable;
+      if (String(variable).match(/^Assistive-/)) {
+        var key = String(variable).replace('Assistive-', '');
+        var element = document.getElementById(key);
+        if (element) {
+          element.value = value;
+        }
       }
     }
   });
--- a/extensions/Assistive-Explore.js
+++ b/extensions/Assistive-Explore.js
@@ -160,7 +160,6 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
     hoverer: null,
     flamer: null,
     speechDiv: null,
-    enriched: {},
     earconFile: location.protocol +
       '//progressiveaccess.com/content/invalid_keypress' +
       (['Firefox', 'Chrome', 'Opera'].indexOf(MathJax.Hub.Browser.name) !== -1 ?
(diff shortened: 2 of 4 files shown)


zorkow:
* Adds better configuration treatment for the explorer.
* Also adds the possibility of adding global configurations for the AT extension. See the Struik file for an example. I am not sure if 'Assistive-' prefix is a good choice there.
* Accessibility menu point is now added before 'About MathJax' 
* Coordinates with Lab.

zorkow:
@dpvc This is now pretty much ready for review.

Semantics-Lab/Lab.js
```diff
@@ -214,9 +214,18 @@ MathJax.Hub.Register.StartupHook("Explorer Ready",
 MathJax.Hub.Register.StartupHook("MathMenu Ready",function () {
   MathJax.Extension.MathMenu.signal.Interest(function (message) {
     if (message[0] === "radio button") {
-      var renderer = message[1].value;
-      if (String(renderer).match(/^(HTML-CSS|CommonHTML|PreviewHTML|NativeMML|SVG)$/)) {
-        Lab.renderer.value = renderer;
+      var value = message[1].value;
+      if (String(value).match(/^(HTML-CSS|CommonHTML|PreviewHTML|NativeMML|SVG)$/)) {
```
dpvc: You are probably right that using `message[1].variable === 'renderer'` here is better.  E.g., replace 217 to 219 with

```
    if (message[1].variable === 'renderer') {
      Lab.renderer.value = message[1].value;
```

extensions/Assistive-Explore.js
```diff
@@ -3,18 +3,62 @@
...
 MathJax.Hub.Register.StartupHook('Sre Ready', function() {
   var FALSE, KEY;
+  var SETTINGS = MathJax.Hub.config.menuSettings;
+
   MathJax.Hub.Register.StartupHook('MathEvents Ready', function() {
     FALSE = MathJax.Extension.MathEvents.Event.False;
     KEY = MathJax.Extension.MathEvents.Event.KEY;
   });
 
+  var Assistive = MathJax.Extension.Assistive = {
```
dpvc: You will want a `version: "1.0"` here, as extensions are supposed to have a version number.  These get listed automatically in the About MathJax dialog box.  The name of the extension is the one you assign here, so your extension will be listed as `Assistive v1.0` (once the version is added).  Is that descriptive enough?

extensions/Assistive-Explore.js
```diff
@@ -3,18 +3,62 @@
...
 
-  var LiveRegion = MathJax.Extension.LiveRegion = MathJax.Object.Subclass({
+    getOption: function(key) {
+      return SETTINGS[Assistive.prefix + key];
+    }
+    
+  };
+  
+  var LiveRegion = MathJax.Object.Subclass({
     version: "1.0",
```
dpvc: Convesely, you don't really need a version here, since this is no longer an extension.  You add it to the Assistive extension as a properly later, though you could do

    var LiveRegion = MathJax.Extension.Assistive.LiveRegion = MathJax.Object.Subclass({

to be consistent with how you handle Explorer below.

extensions/Assistive-Explore.js
```diff
@@ -25,20 +69,56 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
       LiveRegion.Announce();
     },
     //
+    // Shows the live region as a subtitle of a node.
+    //
+    Show: function(node, highlighter) {
+      for (var key in LiveRegion.subtitle) {
+        this.div.style[key] = LiveRegion.subtitle[key];
+      }
```
dpvc: I'm not sure why you don't use a class here and CSS styling to handle the styles for the div.  Wouldn't it be easier just to set its `className` to `MathJax_LiveRegion_Show` or something like that?

If not, you really should make this

```
for (var key in LiveRegion.subtitle) {if (LiveRegion.subtitle.hasOwnProperty(key)) {
  this.div.style[key] = LiveRegion.subtitle[key];
}}
```
so that you're sure that you don't get any other properties in the loop.  For example, if someone adds something to `Object.prototype`, that will end up in your loop.  They aren't supposed to do that, but people do, and we have run into trouble with this in the past, so I've gotten in the habit of always using `hasOwnProperty()`.  (Also, in the old days, some browsers would list ALL properties, here, including built-in methods, and so on; but that was a while ago.)

It is also possible (in modern browsers) to use `Object.keys(LiveRegion.subtitle)` and iterate over that (either with `forEach()`, or with an array counter).

extensions/Assistive-Explore.js
```diff
@@ -25,20 +69,56 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+      this.inner.style.backgroundColor = color.background;
+      this.inner.style.color = color.foreground;
+    },
+    //
+    // Takes the live region out of the page flow.
+    //
+    Hide: function(node) {
+      for (var key in LiveRegion.hidden) {
+        this.div.style[key] = LiveRegion.hidden[key];
+      }
```
dpvc: Again, `hasOwnProperty()` is needed here.

extensions/Assistive-Explore.js
```diff
@@ -25,20 +69,56 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
   }, {
     ANNOUNCE: 'Navigatable Math in page. Explore with shift space.',
     announced: false,
+    hidden: {position: 'absolute', top:'0', height: '1px', width: '1px',
+             padding: '1px', overflow: 'hidden'},
+    subtitle: {position: 'absolute', width: 'auto', height: 'auto',
+               padding: '5px 0px', opacity: 1, 'z-index': '202',
+               left: 0, right: 0, 'margin': '0 auto',
+               backgroundColor: 'white'
+              },
```
dpvc: Should there be a border or drop shadow for the subtitles?

I wasn't able to get them to show up (the box appeared, but not the text), and I could see that a white rectangle was obscuring other text, but it was hard to tell where it started and ended.

Semantics-Lab/Lab.js
zorkow: Done that. Has the effect, that if one chooses to use plain source the combobox is empty. That is probably the behaviour we want anyway.

extensions/Assistive-Explore.js
```diff
@@ -25,20 +69,56 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
     // Speaks a string by poking it into the speech div.
     //
     Update: function(speech) {
-      LiveRegion.Update(this.div, speech);
+      if (Assistive.getOption('speech')) {
+        LiveRegion.Update(this.inner, speech);
+      }
     }
   }, {
     ANNOUNCE: 'Navigatable Math in page. Explore with shift space.',
```
dpvc: Eventually, we will need to make `ANNOUCE` be localizable.  Same for the menus (though that is actually easier), and some other things like the `aria-label`.  That is an unpleasant process, and is a bit harder to third-party extensions (though possible), so perhaps we should wait with it for now and do it when it becomes part of the core.

extensions/Assistive-Explore.js
zorkow: Like look into those tomorrow. Have to go out now.

dpvc: OK, sounds good.

In terms of classes, you can use `MathJax.Ajax.Styles()` to set styles (though it is asynchronous for historical reasons, though it probably doesn't have to be any more).

extensions/Assistive-Explore.js
```diff
@@ -303,7 +368,58 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+    // Attaches the menu;
+    var about = MathJax.Menu.menu.IndexOfId('About');
+    if (about === null) {
+      MathJax.Menu.menu.items.push(ITEM.RULE());
+      MathJax.Menu.menu.items.push(accessibiltyMenu);
+      return;
+    }
+    MathJax.Menu.menu.items.splice(about, 0, ITEM.RULE());
+    MathJax.Menu.menu.items.splice(about, 0, accessibiltyMenu);
+  });
```
dpvc: I'm wondering if the Assistive menu should appear between "Math Settings" and "Language", so that all the "settings" submenus are in the same section?

Here's one way to do that:

```
    var settings = MathJax.Menu.menu.IndexOfId('Settings');
    if (settings == null) settings = MathJax.Menu.menu.length-1;
    MathJax.Menu.menu.items.splice(settings+1, 0, accessibiltyMenu);
```

Note that even if you keep this location, you can do it using a single `splice()` like

```
    var about = MathJax.Menu.menu.IndexOfId('About');
    if (about == null) about = MathJax.Menu.menu.length;
    MathJax.Menu.menu.items.splice(about, 0, accessibiltyMenu, ITEM.RULE());
```

(untested, but the right idea).

extensions/Assistive-Explore.js
```diff
@@ -303,7 +368,58 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
 
-  MathJax.Hub.Register.MessageHook('New Math', ['Register', MathJax.Extension.Explorer]);
+  MathJax.Extension.Assistive.addDefaults();
+
+  MathJax.Hub.Register.StartupHook('MathMenu Ready', function() {
+    var ITEM = MathJax.Menu.ITEM;
+    var accessibiltyMenu =
+          ITEM.SUBMENU(['Accessibility', 'Accessibilty'],
+              ITEM.SUBMENU(['Walker', 'Walker'],
+                  ITEM.RADIO(['dummy', 'Dummy walker'], 'Assistive-walker'),
```
dpvc: Is there a better name that 'Dummy walker' for this menu item?  Like "None" or something?

extensions/Assistive-Explore.js
zorkow: I've been hacking some Python last week.... hence the dreaded for-in.
I know, never use in JS!

extensions/Assistive-Explore.js
zorkow: Unless, I am mistaken, it is not quite the same: In case that About does not exist the accessibility item attaches itself with a rule above. O/w with rule below. 
Collated both calls though.

extensions/Assistive-Explore.js
zorkow: Put "No Walker".

extensions/Assistive-Explore.js
zorkow: Moved everything to classes and added a shadow/border. It should be visible now. E.g. in Struik-speech.html.

zorkow:
Review suggestions incorporated + Bug in Lab.js fixed.
PTAL

zorkow:
@dpvc Hold on with this one. I'll integrate the generateSpeech flag first.

zorkow:
I have now integrated the "generate Speech" flag into the menu. If off, the other speech related menu items will be disabled.
@dpvc In line https://github.com/mathjax/MathJax-RespEq/blob/accessibility_menu/extensions/Assistive-Explore.js#L385 I would like to regenerate the MathML expression for a jax and rerender it in page. What exactly to I need to call to achieve that?

dpvc:
> I would like to regenerate the MathML expression for a jax and rerender it in page.

Do you mean reprocess the input into the internal MathML format?  Or do you mean generate DOM elements from the internal MathML?  Or do you want the serialized MathML from the internal format?  

As for rendering, are you meaning to throw out the current renderer output and reproduce it afresh?  Or do you mean insert the MathML into the page like the AssistiveMML extension?

I'm sorry, I'm just not sure what you are looking for.  Sorry!

It looks like you are trying to do `jax.Reproces()`, which will re-parse the input and re-render the output.  Is that what you want?  You may want to do `MathJax.Hub.Queue(['Reprocess',jax])`.

zorkow:
> Do you mean reprocess the input into the internal MathML format? Or do you mean generate DOM 
> elements from the internal MathML? Or do you want the serialized MathML from the internal format?

Three steps should happen here:
1. a new enriched mathml expression has to be computed with sre
2. the new enriched mml should become the internal format of the jax
3. the jax should be rendered in page, so that the changed data-attributes are also in the page.

zorkow:
* I have the desired functionality on "generate speech" change.
* I've removed all caching.

If there are no major problems any more I suggest we merge and continue with some of the other things in PR #78. Also I have a branch based on this with a potential solution for issue #67.

dpvc:
> Three steps should happen here...

All three should be obtained by `MathJax.Hub.Queue(['Reprocess',jax])`, I think.  (I didn't test it, but that is what should happen in theory).

> I suggest we merge and continue with some of the other things in PR 78.

I agree, and was going to suggest something like that.  If you are ready to merge, then go ahead.

zorkow:
Indeed, it's already in the code and worked fine.

Issue #66, opened 2016-02-01
[a11y walker] enable inline math


pkra:
Currently, all extensions skip inline math. That makes sense for responsive rendering (though I might argue for that there, too). But for the a11y walker we need to enable inline math elements.

dpvc:
The `Semantic-Collapse` extension only collapses displayed equations, but `Semantic-MathML` should process all math, and I don't see anything in the walker that would exclude in-line math.  In fact, I've just verified that the walked *does* walk in-line math.

If you could post the example file where this isn't working for you, that might help.

pkra:
Hm. Maybe I was just sleep deprived after the deadline. I'll check.

zorkow:
@dpvc I am trying to understand why the inline math expressions do not get the data-semantic attributes attached and am a bit a loss. I am working with the Struik example.

They are being enriched correctly, with jax.enriched  set to the enriched MathML. They also get the correct events added, but the data-semantic attributes are not pushed through, so we can not speak or walk. Is there another round of rerendering missing?

zorkow:
FWIW, I've run reprocess() but that does not help either.

pkra:
> FWIW, I've run reprocess() but that does not help either.

Have you tried rerender? IIRC, reprocess re-digests the input which (IIUC) wouldn't have changed.

zorkow:
Tried both.

pkra:
Ah, I should've figured...

dpvc:
It looks like some of the Semantic-Collapse code needs to be refactored into Semantic-MathML, and that may mean that the pipe-line between the two needs improvement.  I'll have to see what can be done.

As a temporary work-around, you could probably add

```
}
if (jax.enriched) {
```

right after [line 156](https://github.com/mathjax/MathJax-RespEq/blob/master/extensions/Semantic-Collapse.js#L156) in Semantic-Collapse.js so that the enriched form is always turned into the internal form (this was only being done for display equations, as Peter pointed out; not sure what walking I was doing, perhaps by syntax rather than semantics; sorry about that).

zorkow:
PR #77 contains the work around.

dpvc:
PR #79 should resolve this issue.

zorkow:
Done.

PR #77, opened 2016-02-13
Enriches Inline expressions

zorkow:
* Inline expressions now also use the enriched MathML as internal format.
* They get mactions and can be collapsed manually, but are not collapsed automatically. 
* A version of the Struik example (Struick-speech.html) demonstrates the effect.

I still feel that this logic should be refactored to Semantic-MathML. I.e., the internal format should be reset in the Filter method there and the MakeMML method should be moved. I am just not quite sure what the AddChildren method does, that uses MakeMML as well.

@dpvc Any objections?

### DIFF
--- a/examples/Struik-speech.html
+++ b/examples/Struik-speech.html
@@ -1,0 +1,191 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<meta http-equiv="X-UA-Compatible" content="IE=edge" />
+<title>MathJax:  Collapsing Differential Geometry Examples</title>
+
+<script type="text/x-mathjax-config">
+MathJax.Hub.Config({
+  tex2jax: {inlineMath: [['$','$'],['\\(','\\)']]},
+  SemanticCollapse: {autoCollapse: true},
+  AssistiveMML: {disabled: true},
+});
+MathJax.Ajax.config.path["RespEq"] = "../extensions"
+MathJax.Hub.config.extensions.push(
+"[RespEq]/Assistive-Explore.js",
+"[RespEq]/Semantic-Collapse.js");
+MathJax.Hub.Register.StartupHook('Explorer Ready', function()
+{MathJax.Extension.Explorer.config.walker = "syntactic";});
+</script>
+<script src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML-full"></script>
+</head>
+<body>
+
+<blockquote>
+
+<h1>An example from Struik's <i>Lectures on Classical Differential Geometry</i></h1>
+
+<!------------------------------------------------------------------------>
+<hr>
+<div id="p23">
+p. 23
+<p>
+
+<blockquote>
+
+<p>
+This procedure is simply a generalization of the method used in Sects.
+1-3 and 1-4 to obtain the equations of the osculating plane and the
+osculating circle.  Let $f(u)$ near $P(u=u_0)$ have finite derivatives
+$f^{(i)}(u_0)$, $i = 1, 2, \ldots, n+1$.  Then if we take $u=u_1$ at $A$
+and write $h = u_1 - u_0$, then there exists a Taylor development of $f(u)$
+of the form (compare Eq.  (1-5)):
+$$
+f(u_1) = f(u_0) + hf'(u_0)+{h^2\over 2!}f''(u_0) + \cdots
+ + {h^{n+1}\over (n+1)!}f^{(n+1)}(u_0) + o(h^{n+1}).
+$$
+
+Here, $f(u_0)=0$ since $P$ lies on $\Sigma_2$, and $h$ is of order $AP$
+(see theorem Sec.  1-2); $f(u_1)$ is of order $AD$.  <i>Hence necessary and
+sufficient conditions that the surface has a contact of order $n$ at $P$
+with the curve are that at $P$ the relations hold:</i>
+
+$$
+f(u) = f'(u) = f''(u) = \cdots = f^{(n)}(u) = 0;\quad f^{(n+1)}(u) \ne 0.
+$$
+
+</blockquote>
+
+<!------------------------------------------------------------------------>
+<hr>
+<div id="p40">
+p. 40
+<p>
+
+<blockquote>
+
+The converse problem is somewhat more complicated: Find the curves which
+admit a given curve $C$ as involute.  Such curves are called
+<i>evolutes</i> of $C$ (German: <i>Evolute;</i> French:
+<i>d&eacute;velopp&eacute;es</i>).  Their tangents are normal to $C({\bf
+x})$ and we can therefore write the equation of the evolute ${\bf y}$ (Fig.
+1-34):
+
+$$
+{\bf y} = {\bf x} + a_1{\bf n} + a_2{\bf b}.
+$$
+
+Hence
+
+$$
+{d{\bf y}\over ds} = {\bf t}(1-a_1\kappa) + {\bf n}\left({da_1\over
+ds}-\tau a_2\right) + {\bf b}\left({da_2\over ds}+\tau a_1\right)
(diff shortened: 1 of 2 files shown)


extensions/Semantic-Collapse.js
```diff
@@ -146,16 +146,16 @@
...
-              script.parentNode.childNodes.length <= 3)) {
-        if (jax.enriched.nodeName.toLowerCase() !== "math") {
-          var math = document.createElement("math");
-          math.appendChild(jax.enriched);
-          jax.enriched = math;
-        }
-        jax.root = this.MakeMML(jax.enriched);
-        jax.root.inputID = script.id;
+      if (!jax.enriched) { return; }
+      if (jax.enriched.nodeName.toLowerCase() !== "math") {
```
dpvc: This `if` block probably isn't needed any more.  At one point SRE could return a non-math node at the top level, but I think that has been fixed.

zorkow: Oh, yes I vaguely remember that bug. So effectively jax.enriched does not change between line
Semantic-MathML [line 97](https://github.com/mathjax/MathJax-RespEq/blob/master/extensions/Semantic-MathML.js#L97) and here?

dpvc:
> I still feel that this logic should be refactored to Semantic-MathML.

I agree with that, as I mention [yesterday](https://github.com/mathjax/MathJax-RespEq/issues/66#issuecomment-183441112).  I just haven't had the chance to do it yet.

The reason it is this way is that the adding of the adding of the `maction` nodes for the collapsing is being done at the same time the serialized MathML from the enrichment process is being parsed.  To break the two actions apart, the Semantic-MathML would need to do the parsed into the internal javascript version of the MathML, and then the Semantic-Collapse will need to walk the tree and modify it to add the `maction` elements and the complexity data after the fact (or re-serialize and re-parse, but that is a waste, and is why they were combined originally).

zorkow:
Get it. So effectively, what we would like to have are three modules: 
Semantic-MathML: Just adds the enriched MML, which is sufficient for walking and speaking.
Semantic-Action: Adds actions and complexity. That is sufficient for flame/hover highlighting.
Semantic-Collapse: Executes the collapse on page load.

Is that roughly correct?

dpvc:
> the internal format should be reset in the Filter method there and the MakeMML method should be moved.

The `MakeMML()` method is the one that currently is handling the computation of complexity and addition of `maction` elements.  This, and the `AddChildren()` and a couple other methods are (simplified) versions of the ones from the MathML input jax.  These recursively compute the complexity and add in the `maction` nodes based on that.

If the enriched MathML is to be parsed in Semantic-MathML, there is no need for this modified `MakeMML()`, as you can just use the one from the MathML input jax directly (it will have to be forced to load in case a different input method is being used, though).  Then Semantic-Collapse.js will have to be modified to walk the tree rather than work through the parsing process, as is currently done.

dpvc:
> Is that roughly correct?

That sounds reasonable.  I hadn't planned to separate the Semantic-Action from Semantic-Collapse, but that makes sense.

extensions/Semantic-Collapse.js
dpvc: The `jax.enriched` property is the string returned from SRE at [line 28](https://github.com/mathjax/MathJax-RespEq/blob/master/extensions/Semantic-MathML.js#L28), which is never changed once it is assigned (unless the jax is re-processed).  If the parsing of the enriched form is moved to `Semantic-MathML.js`, then this should be parsed immediately after line 28, and there is no need to keep thee string attached to the jax.

examples/Struik-speech.html
```diff
@@ -0,0 +1,191 @@
...
+  tex2jax: {inlineMath: [['$','$'],['\\(','\\)']]},
+  SemanticCollapse: {autoCollapse: true},
+  AssistiveMML: {disabled: true},
+});
+MathJax.Ajax.config.path["RespEq"] = "../extensions"
+MathJax.Hub.config.extensions.push(
+"[RespEq]/Assistive-Explore.js",
+"[RespEq]/Semantic-Collapse.js");
+MathJax.Hub.Register.StartupHook('Explorer Ready', function()
+{MathJax.Extension.Explorer.config.walker = "syntactic";});
```
dpvc: We should make the explorer extension be configurable via the `MathJax.Hub.Config()` mechanism, so that you could do

```
MathJax.Hub.Config({
  tex2jax: {inlineMath: [['$','$'],['\\(','\\)']]},
  SemanticCollapse: {autoCollapse: true},
  AssistiveMML: {disabled: true},
  Explorer: {walker: "syntactic"}
});
```

rather than having to do a startup hook for this.

examples/Struik-speech.html
```diff
@@ -0,0 +1,191 @@
...
+<head>
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<meta http-equiv="X-UA-Compatible" content="IE=edge" />
+<title>MathJax:  Collapsing Differential Geometry Examples</title>
+
+<script type="text/x-mathjax-config">
+MathJax.Hub.Config({
+  tex2jax: {inlineMath: [['$','$'],['\\(','\\)']]},
+  SemanticCollapse: {autoCollapse: true},
+  AssistiveMML: {disabled: true},
```
dpvc: The final comma should be removed (it crashes IE < 9, though I guess we don't care about that anymore).

examples/Struik-speech.html
zorkow: Have a look at #76. The explorer options can now be set via menuSettings.
We could of course move those out. 
But this discussion should be had for #76 as this is more "modern".

zorkow:
We would need to keep the jax.enriched around, as the explorer depends on it.
However, a simple flag would be enough rather than the whole string.

zorkow:
I've been playing with your idea of using the original MathML parsing. But we would then have to wait for the MathML input jax to be loaded.

examples/Struik-speech.html
zorkow: True, but no need to be sloppy. But the Struik-speech file is ephemeral anyway.

examples/Struik-speech.html
dpvc: OK.  I'm going through that as we speak.

So then should this call be replaced by

```
MathJax.Hub.Config({
  tex2jax: {inlineMath: [['$','$'],['\\(','\\)']]},
  SemanticCollapse: {autoCollapse: true},
  AssistiveMML: {disabled: true},
  menuSettings: {"Assitive-walker": "syntactic"}
});
```

instead?

zorkow: Yes, once #76 is merged and provided we keep this file around. Then the entire load sequence would be simplified. (See the Struik.html in #76).

dpvc:
> we would then have to wait for the MathML input jax to be loaded.

Correct.  But without it, we would need to duplicate that code.  I didn't really like doing that in `Semantic-Collapse.js`, and only did it because of the addition of the complexity computations (and tried several ways to avoid it).  In `Semantic-MathML.js`, it would be pure duplication.

examples/Struik-speech.html
dpvc: Right.  I had seen that and then forgotten it.

dpvc:
OK, I will merge.

zorkow:
I agree, we should not duplicate that code. But I could not see a signal from the MathML input jax. Does it post one?

dpvc:
> But I could not see a signal from the MathML input jax. Does it post one?

Yes, the jax signals are posted automatically by the base jax class when they are loaded.  You can look for `"MathML Jax Ready"`.

dpvc:
You can do something like

```
MathJax.Ajax.Require("[MathJax]/jax/input/MathML/config.js",
     ["Require",MathJax.Ajax,"[MathJax]/jax/input/MathML/.js"]);
MathJax.Hub.Register.StartupHook("MathML Jax Ready",function () {...})
```

to load the two pieces needed for the MathML jax, and wait for it to load.

PR #78, opened 2016-02-14
Some adjustments for Voker to consider.

dpvc:
Here are some minor things that I saw looking at the code.

* The URL for the earcon should be based on the `MathJax.Ajax.config.path` used for SRE, since that can be set by the page author.  (In particular, I use it to run a local copy of SRE during testing.)  In any case, if you use that here, it will be easier to move this to the third-party repository or the core MathJax directories in the future.

* There is no need for each math element to create a new function closure for the event listener, since it is not dependent on the particular math element.  So I've made a common one that gets used for all.

* The `FALSE` function from `MathJax.Extensions.MathEvents` should be returned by the event handler, not be run followed by a return of null.

* Since you have `FALSE` defined, I've substituted that for some direct calls to `MathJax.Extensions.MathEvents.False()`.

* There is no need for the `return` at the bottom of the `Flame()` method.

There are also a couple of other issues that I wanted to discuss.

*  There are a couple of potential problems using `Explorer.enriched` to cache the DOM elements.  The original reason that we never cached pointers to DOM objects that contain pointers back into the javascript (which the `<script>` elements actually do) is that this could cause circular references that produced memory leaks in some browsers (notably IE).  Microsoft claims that they have eliminated this memory leak, but I haven't verified that, and I don't particularly trust them.  This is why the internal MathML objets only have ID's into the DOM and look them up with `getElementById()` when needed, rather than actually caching pointers to the DOM elements created by the output renderers (even though that would have been very convenient).

    That issue may be resolved now, but the other problem is that the math on the page may not be constant.  Users create and destroy elements all the time, and an extreme case is an editor like in StackExchange where every keystroke causes the math to be removed and rebuilt.  If you are caching the math, you will end up with lots of math that has been removed from the DOM.  This will mean that those elements will not be able to be freed (because of your pointers), and the code that uses those pointers may end up looking at elements that are no longer in the DOM.

    I would recommend not keeping a cache at all.  Instead, use `MathJax.Hub.getAllJax()` and run through them for the ones that are marks with `enriched` (or use your own flag).  That is safer and you will always end up with only the elements actually in the DOM (and won't end up growing the cache as the page changes repeatedly).

* In `AddEvent` [line 212](https://github.com/mathjax/MathJax-RespEq/blob/87a48dd6ee227e6f681ff7d446c2d1c60b5c85b5/extensions/Assistive-Explore.js#L212), you queue another call to `AddEvent`.  I'm concerned that this could lead to an infinite loop, and am not sure of the use case you have in mind.  I'm assuming that it is for the case where the output is not yet in place for the math (the previous sibling isn't what you expect), but I'm not sure of when that would occur.  Do you have a situation where this was needed in practice?

* If you retain the cached `Explorer.enriched`, then at [line 304](https://github.com/mathjax/MathJax-RespEq/blob/87a48dd6ee227e6f681ff7d446c2d1c60b5c85b5/extensions/Assistive-Explore.js#L304) in `FlameEnriched`, then I think the `previousSibling` may be null (if the DOM elements have been removed from the page).  It doesn't look like you check for that, and may want to (you might even want to remove the element from the cache in that case).

*  If `generateSpeech` is false (as in the Struick example), then I could not get subtitles to show up (they are just blank).  I didn't try it with a screen reader, but I assume that means no speech either.  In this case, the menu items for "Speech Output" and "Subtitles" seem to have no effect.  I think that either the viewer's menu settings should override the author's (my personal preference), or the menu items should be cleared and disabled if `generateSpeech` is false.

    I'm not really sure why there is a separate value for this; doesn't "Speech Output" cover that?  Or is that something different?  Also, if `generateSpeech` is no controlled by the menu, then it should not be stored in `menuSettings`, but should be part of the general configuration for the Assistive extension.  (There probably does need to be an Assistive configuration block for things like that, and the `ANNOUNCE` value, the `aria-label`, etc.)

*  Finally, I could not get the Flame highlighting to work.  Hover worked, but not Flame.  Perhaps I wasn't using it right.


zorkow:
... Volker ...

I agree with most item 1,2,4,5. Item 3: No! See inline comments in code.

Item 6: Since register is called on each New Math signal, I would assume the cached versions will be overwritten if they dynamically change. I agree that if an item vanishes altogether, than it remains in the cache, which could lead to problems later on.

Item 7: The AddEvent seems to be a leftover from before I fully understood the signalling and queuing in MathJax. (I am not claiming that I have fully understood now!)

Item 8: I am not sure I understand the problem here.

Item 9: generateSpeech was intended for us to switch off speech generation quickly. In the normal use case speech should always be generated. However, I could see some use for this and it might be a good idea to integrate it properly with the other options and the menu.

Item 10: I'll look into this flaming problem!

zorkow:
Item 10: Flame works fine in the semantic lab but not in the Struik example. The reason is that the maction elements are given (as classes) in the rendered expression in the Lab but not in Struik.
The reason for this, however, I can't see. @dpvc Any idea?

dpvc:
> Any idea? [about flame]

I'll have to look into it.  I don't have any suggestions yet.

> I would assume the cached versions will be overwritten if they dynamically change.

That depends on how the page author changes the math.  If he or she calls the `jax.Text()` method to change the original input text for the jax, then yes.  But this is rarely used.  More likely, they will set `innerHTML` on some container and replace the math altogether.  In that case, the math will get a new ID and you will have outdated IDs in your cache.

> Item 8: I am not sure I understand the problem here.

If an author replaces the `innerHTML`, for example, and so the math that you have cached is no longer in the DOM, there are two consequences.  First, the DOM element can not be de-allocated and garbage collected since you have an active pointer to it, so there is memory that won't be freed until the viewer leaves the page.  Second, since the DOM element is no longer in the page, its `parentNode`, `previousSibling`, `nextSibling`, and other such pointers will be `null` (I'm pretty sure only the one DOM node will be retained and the others will be freed), so when you use the cached element's `previousSibling`, that will be `null`.  I don't know if your `Flame()` routine handles that or not (since it is used in `Explorer.highlighter.highlightAll(node);`, and I haven't checked if that handles a null `node` or not).

PR #80, opened 2016-03-02
Handle the default key names properly.

dpvc:
This fixes the problem with Struik and the walker, where the defaults were not being set properly, so the colors for foreground and background weren't getting set.

The problem was that `Object.keys()` returns an array, so `var keys in Object.keys(...)` was returning indices rather than the names themselves.  This fixes it to get the array of keys and then loop through the indices.


zorkow:
lgtm.

PR #81, opened 2016-03-02
Fix timing problem with refactored code

dpvc:
This pull request moves `loadComplete()` calls to after the extensions are ready (to get proper synchronization with the startup sequence).  Fixes the timing problem that Volker and Peter were experiencing.


zorkow:
lgtm.

PR #82, opened 2016-03-03
Add labelledby element

zorkow:
Adds a hidden labelledby element into the DOM for  each Maths expression to be spoken in continuous read mode by some SRS.

### DIFF
--- a/extensions/Assistive-Explore.js
+++ b/extensions/Assistive-Explore.js
@@ -207,6 +207,21 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
         }
         if (math) {
           math.onkeydown = Explorer.Keydown;
+          //
+          var speechGenerator = new sre.DirectSpeechGenerator();
+          var span = math.querySelector('.math');
+          if (span) {
+            var speech = speechGenerator.getSpeech(span);
+            if (speech) {
+              var descr = id + '-Descr';
+              math.setAttribute('aria-labelledby', descr);
+              var newSpan = MathJax.HTML.addElement(
+                document.body, 'div',
+                {id: descr, className: 'MJX_LiveRegion'});
+              newSpan.textContent = speech;
+            }
+          }
+          //
           math.addEventListener(
             Explorer.focusEvent,
             function(event) {
--- a/extensions/Assistive-Explore.js
+++ b/extensions/Assistive-Explore.js
@@ -119,7 +119,13 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
                left: 0, right: 0, 'margin': '0 auto',
                'background-color': 'white', 'box-shadow': '0px 10px 20px #888',
                border: '2px solid #CCCCCC'
-             }},
+             },
+             '.MJX_DescriptionElement':
+             {
+               position: 'absolute', top:'0', height: '1px', width: '1px',
+               padding: '1px', overflow: 'hidden'
+             }
+            },
     //
     // Creates a live region with a particular type.
     //
@@ -210,7 +216,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
           math.onkeydown = Explorer.Keydown;
           //
           var speechGenerator = new sre.DirectSpeechGenerator();
-          var span = math.querySelector('.math');
+          var span = math.querySelector('[data-semantic-speech]');
           if (span) {
             var speech = speechGenerator.getSpeech(span);
             if (speech) {
@@ -218,7 +224,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
               math.setAttribute('aria-labelledby', descr);
               var newSpan = MathJax.HTML.addElement(
                 document.body, 'div',
-                {id: descr, className: 'MJX_LiveRegion'});
+                {id: descr, className: 'MJX_DescriptionElement'});
               newSpan.textContent = speech;
             }
           }


extensions/Assistive-Explore.js
```diff
@@ -208,6 +208,21 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
         }
         if (math) {
           math.onkeydown = Explorer.Keydown;
+          //
+          var speechGenerator = new sre.DirectSpeechGenerator();
+          var span = math.querySelector('.math');
```
dpvc: Currently, this only finds HTML-CSS output.  If you use

          var span = math.querySelector('.math, .mjx-math, .mjx-svg-math, .MJXp-math, math');

instead, it should get everything except PlainSource output.  (Yes, these should be consolidated for v3.0).

extensions/Assistive-Explore.js
```diff
@@ -208,6 +208,21 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+          if (span) {
+            var speech = speechGenerator.getSpeech(span);
+            if (speech) {
+              var descr = id + '-Descr';
+              math.setAttribute('aria-labelledby', descr);
+              var newSpan = MathJax.HTML.addElement(
+                document.body, 'div',
+                {id: descr, className: 'MJX_LiveRegion'});
```
dpvc: This isn't really a live region, is it?  So shouldn't it have a different class name?

Also, won't these divs all be read when the reader reaches the bottom of the page (or does the fact that it is the target of an `aria-labelledby` prevent that)?  So they don't need to be marked in any way to prevent that?

Is there any chance that the content of these divs will ever look like TeX or AsciiMath?  That is, will there ever be a raw dollar sign or `\begin` or a back tick in the string?  If so, then there would be the possibility, if `MathJax.Hub.Typeset()` were run a second time (as happens with pages that add content after the initial typeset pass), that MathJax would try to typeset math *inside* one of these.  Even of not, MathJax has to scan these for math in that case, and on pages with lots of math, that could be a lot of extra scanning.

Would it be possible to add `tex2jax_ignore` and `asciimath2jax_ignore` classes to the list for these (so they will be skipped)?  Or perhaps put them all into a container that has these two classes?

extensions/Assistive-Explore.js
zorkow: Even better. I replace it with math.querySelector('[data-semantic-speech]');
That makes sure we catch the first element that actually contains speech, which is not always the top element. (had forgotten about that).

extensions/Assistive-Explore.js
zorkow: > This isn't really a live region, is it? So shouldn't it have a different class name?

I made it a liveregion so it would be outside the page flow, similar to the other live regions. That's what prevents them to be spoken at the end of the page. 
But we can give it a different class name and give that just the same style as the MJX_LiveRegion elements. This way we can control them separately.

> Is there any chance that the content of these divs will ever look like TeX or AsciiMath?

They should not. At lest with the current speech generators. Is there a change of double quote ever being a delimiter?

dpvc: > Is there a change of double quote ever being a delimiter?

Since people get to set their own delimiters, I guess the answer is "yes", but it is a slim chance.

dpvc: Since the page author can change the tex2jax and asciimath2jax ignored classes anyway, perhaps the best thing would be for me to modify those preprocessors to have some fixed classes to ignore in addition to the ones from the page author.  So I would not worry about this for now.

dpvc:
I'm merging this for now.  We can deal with the ignore-class stuff later.

zorkow:
> > Is there a change of double quote ever being a delimiter?
> Since people get to set their own delimiters, I guess the answer is "yes", but it is a slim chance.

But it should still not matter, since the quotes are just the string quotes of the textContent element which I understand are not really part of the DOM. Otherwise (at least for the mathspeak rules) it's only alphanum and dash in the speech strings. So that should not be problematic.
Of course authors can add/change speech rules... but that might take a while until anyone does.

extensions/Assistive-Explore.js
dpvc: > I made it a liveregion so it would be outside the page flow, similar to the other live regions. That's what prevents them to be spoken at the end of the page.

I know next to nothing about ARIA, so I may be off, here, but don't you need something like `aria-live` to make it a live region?  The CSS makes it not visible and does remove it from the main flow, but it is still part of the page, it I think it still gets read.  I just tested in VoiceOver, and it does get read when you get to the end of the page.  I haven't tested with anything else.

This CSS was used with things like the assistive MathML so that it would not be displayed to the visual user, but would still be read by the screen reader.  It was specifically designed *not* to be invisible to the screen reader.  Am I missing something obvious?

zorkow: Sorry, that was very badly phrased. I gave it the class MJX_LiveRegion, so it would pick up the right style to be outside the page flow.

dpvc:
>  the quotes are just the string quotes of the textContent element which I understand are not really part of the DOM

I'm not sure what you mean by "not part of the DOM", because setting the `textContent` field deletes any child nodes and then adds a text node containing the text.  So there is definitely a text node in the DOM with the specified text.  I checked that to make sure.

That means that things like `tex2jax` will scan that text, and if the delimiters are set to things in the string, MathJax *will* process them.  I checked that, too, and it really does happen.

Of course, the chance that the delimiters will be set to plain words is very small.

extensions/Assistive-Explore.js
zorkow: I had tested with VoiceOver before and it did not get read. However, that was before we updated to ElCapitan. If it gets read, that's worrying.
Today I only tested with ChromeVox and it did not get read.

zorkow:
I mean the double quotes are not part of the DOM, as in given explicitly. 
The textContent is of course part of the DOM. E.g., if you inspect a text node in Chrome it looks for example like <p> "blurb" </p> , but in FF it would look like <p>blurb</p>

extensions/Assistive-Explore.js
zorkow: @pkra did some testing with the labelledby nodes. Were there any problems in NVDA for example?

zorkow: @dpvc What happens if you add visibility hidden? Does VoiceOver still pick it up?

dpvc:
One thing that I noticed that *is* an issue, however, is that if existing math is re-typeset (e.g., via `MathJax.Hub.Rerender()`, or by using an jax's `jax.Text()` method to change the math expression), a second  div with the same `id` will be created, and the old one will remain in place.  This makes for invalid HTML, and the first element takes precedence over the second, so the label that is found will be the old value rather than the new one.  It could also lead to the page growing larger and larger if this is done more than once.

A related issue is that `jax.Remove()`, which is supposed to remove the output for the jax, doesn't remove this label div.  Also, if the page dynamically changes the content so that the math is removed, these divs will still be in the page.  They are hidden, so it is not crucial, but in a situation like an editor preview (which replaces the preview on every keystroke and serenaders the contents), this could lead to *lots* of extra detached labels.

One thing that would help would be to check if an element with the given id already exists, and reuse it rather than create a new one.  That would take care of the re-render problem.

Another thing that would help would be to place the label div within the element's `-Frame` span rather than the end of the document, like is done with the assistive MathML output.  That would mean that if the math is removed from the page (via the page setting `innerHTML` for a container element of the math, for example), then these divs will be remove along with the math and its output.

To take care of `jax.Remove()`, you would have to override the jax `Remove()` method with a new method that calls the old one, and then clean up any additional DOM elements that you have added.  That is awkward, but would do the trick.  But `Remove()` is not used very much, so it may not be worth it.

In 3.0, maybe there should be a means of registering DOM elements that would need cleaning up.  (This is not a use case that we considered in the original design.)  Alternatively, I have been planning to have a container element for each expression that contains all the nodes used by MathJax for that expression (something like the `-Frame` element, but that doesn't currently contain the script or preview elements; I want something that contains everything used by MathJax for that expression, so it is easier to locate that stuff, and to remove it.  Currently there are complicated and fragile routines for doing that.

extensions/Assistive-Explore.js
dpvc: It doesn't seem to read it then.  I'm running an older system on my laptop.  Let me try an El Capitan version and see what happens there...

dpvc: I tested in El Capitan, and it also reads the description divs at the bottom of the page, bit doesn't if the visibility is set to hidden.

On the other hand, I can't seem to get VoiceOver to read the labeled-by element at all in El Capitan; it did in the earlier version, but in that case, it read it for every node in the HTML-CSS output for the math.

Of course, I don't know much about using VoiceOver, so there may be modes where it works better.  I used the "Read all" feature to read the whole page (ending divs were read), and the "Quick Nav" feature to move around the page by hand (end divs were included in the navigation).

PR #83, opened 2016-03-03
Fixes the issue that flaming would only work on one element.

zorkow:
Classic example where a compiler would help. Closure would immediately flag a warning unused parameter 'node' .

### DIFF
--- a/extensions/Assistive-Explore.js
+++ b/extensions/Assistive-Explore.js
@@ -313,7 +313,6 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
     // Activates Flaming
     //
     Flame: function(node) {
-      Explorer.UnFlame(node);
       if (Assistive.getOption('highlight') === 'flame') {
         Explorer.GetHighlighter(.05);
         Explorer.highlighter.highlightAll(node);
@@ -321,13 +320,14 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
         return;
       }
     },
-    UnFlame: function(node) {
+    UnFlame: function() {
       if (Explorer.flamer) {
         Explorer.highlighter.unhighlightAll();
         Explorer.flamer = null;
       }
     },
     FlameEnriched: function() {
+      Explorer.UnFlame();
       for (var i = 0, all = MathJax.Hub.getAllJax(), jax; jax = all[i]; i++) {
         Explorer.Flame(document.getElementById(jax.inputID).previousSibling);
       }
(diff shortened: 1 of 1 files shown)


dpvc:
This looks better, but I'm still having one problem.  If I resize the Struik page (so that the automatic collapse occurs), then the flame is not applied to the redrawn equation.  You might need to tie into the "New Math" signal to add the flaming when it is selected and new math arrives.

This also shows up in the semantics lab pages, if you turn on flame and then change the equation and typeset it.

zorkow:
It also does not work when flame is switched on on page load. E.g., if it is set in the menu/cookie.

dpvc:
> It also does not work when flame is switched on on page load.

Right.  I think that is the same solution: check for flame being needed during the "New Math" handler (the `Register()` function).

zorkow:
"End Math" actually does the trick.

zorkow:
That should work now. ptal.

I found a couple of issues related to walking on collapsed equations, but it also gave me an idea how we could finally integrate the summary speech without disrupting the current process flow.

dpvc:
> "End Math" actually does the trick.

This looks like it removes all the flaming and then adds it in again.  That seems inefficient in the case where new math has been added to the page and typeset (all the existing math is unchanged, so doesn't need to have flaming removed and re-added), or where one expression has changed (due to `jax.Text()` call).  Some pages have hundreds or thousands of expressions on them, so removing and reading the flames can be non-trivial.

zorkow:
That's true. Reset is the function used when changing parameters like fore/background.
But if that is the correct signal, then I can write a node based method.

dpvc:
I think you already have one:  `Explorer.Flame(node)`.  Can't you just add

```
if (Explorer.flamer) Explorer.Flame(script.previousSibling);
```

after [line 191](https://github.com/mathjax/MathJax-RespEq/blob/flame_issue/extensions/Assistive-Explore.js#L185-L194)?

zorkow:
Just making sure I understand: After a jax.Text() I have a fresh  expression? I.e., no old highlighting bits linger around anymore?

zorkow:
Yes effectively, that is the function. I still need to extract the node from the message, though.

dpvc:
> After a jax.Text() I have a fresh expression? I.e., no old highlighting bits linger around anymore?

Yes (I haven't tested, but I believe that is correct).

> I still need to extract the node from the message, though.

You have already gotten the script node, and so I think you already did that.  It is the previous sibling from that that you need (according to [this line](https://github.com/mathjax/MathJax-RespEq/blob/flame_issue/extensions/Assistive-Explore.js#L332)).

Just FYI, there is API for getting the script element from the jax:  `jax.SourceElement()`, so you don't have to do the `document.getElementById(jax.inputID)` yourself.

zorkow:
```jax.SourceElement()``` is a useful function.
It appears that there are no residue highlights. At least from what could see and test.

dpvc:
OK, that looks good, and resolves the issues with re-rendering the math and reloading the page.  I'll merge.

dpvc:
> `jax.SourceElement()` is a useful function.

Every now and again, I get one right.  :-)

PR #85, opened 2016-03-07
Adapt struik speech

zorkow:
Fixes the Struik-speech file for user testing.
@pkra This is the one we might want to send out.

### DIFF
--- a/examples/Struik-speech.html
+++ b/examples/Struik-speech.html
@@ -8,9 +8,13 @@
 <script type="text/x-mathjax-config">
 MathJax.Hub.Config({
   tex2jax: {inlineMath: [['$','$'],['\\(','\\)']]},
-  SemanticCollapse: {autoCollapse: true},
+  "fast-preview": {disabled: true},
+  SemanticCollapse: {disabled: true},
   AssistiveMML: {disabled: true},
-  menuSettings: {'Assistive-walker': 'syntactic'}
+  menuSettings: {
+    'Assistive-generateSpeech': true,
+    'Assistive-walker': 'syntactic'
+  }
 });
 MathJax.Ajax.config.path["RespEq"] = "../extensions"
 MathJax.Hub.config.extensions.push(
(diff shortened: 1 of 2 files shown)


pkra:
Checking it out now.

pkra:
@zorkow are processing errors on IE10 expected? Last I checked SRE should work?

examples/Struik-speech.html
```diff
@@ -8,9 +8,14 @@
 <script type="text/x-mathjax-config">
 MathJax.Hub.Config({
   tex2jax: {inlineMath: [['$','$'],['\\(','\\)']]},
-  SemanticCollapse: {autoCollapse: true},
+  "fast-preview": {disabled: true},
+  SemanticCollapse: {disabled: true},
+  SemanticComplexity: {disabled: true},
```
dpvc: I think the better way to do this is to only load `Semantic-Math` rather than `Semantic-Collapse` below.  There is need to load the code you are not intending to use.

zorkow:
I can't see any processing errors in IE10.

zorkow:
@pkra I can't see the processing error. But I get a null reference error in Semantic-Collapse.js

zorkow:
OK, I've tested extensively with IE11 (Win7) and IE10(Win7). Everything works fine with IE11.
IE 10 is a different beast...
Here are some observations (we might know some already), unless otherwise stated, they are for both Struik and Struik-speech (in the latter collapse is off):
* HTML-CSS renders out of sync. 
* When collapse is on, those equations that can be collapsed jump in there correct place, once they have been collapsed once.
* They do not have walker events attached, so can't be explored. This suggests that something goes wrong before the New Math signal is sent.
* Reprocess crashes the page.
* CommonHTML and SVG renderer work fine.
* For all renderers there are three math processing errors on the page. 
* Renderers can not be changed without page reload.

I checked, but everything seems to be coming back fine from SRE, so I don't think the problem is there. It is sometimes, however, not even being called.

@pkra When was the last time IE10 worked? Was that before the refactoring of Semantic-* into three modules? Did you test with IE10 today? If so, it must be an SRE problem, but I can't really see where.

pkra:
> Did you test with IE10 today?

Yes. It mostly works but some equations would have Math Processing Errors (further down in Struik).

PR #92, opened 2016-04-19
Abstraction rules

zorkow:
This is the final integration the latest developments:
* Enables collapse and expansion with keyboard.
* Integrates speech for collapsed elements.
* Retains state of exploration for each element.
* Refactors announcement to first focus from page load. 
* Allows choice of speech rules and styles.

Still two things are missing:
* Actionable nodes (mactions) are not announced.
* The subtitles can sometimes overlap expressions after a collapse/expand action.


extensions/Assistive-Explore.js
```diff
@@ -163,16 +183,19 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
     flamer: null,
     speechDiv: null,
     earconFile: location.protocol +
-      '//progressiveaccess.com/content/invalid_keypress' +
+      '//mathjax.github.io/MathJax-accessibility/dist/invalid_keypress' +
```
dpvc: if `location.protocol` is `file:` then this is not going to get you the correct location.  Why not just use `https:` all the time?

extensions/Assistive-Explore.js
```diff
@@ -195,46 +228,108 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
-              math.setAttribute('aria-label', speech);
+      if (!sibling) return;
+      var math = sibling.id !== id ? sibling.firstElementChild : sibling;
+      Explorer.AddAria(math);
+      Explorer.AddMouseEvents(math);
+      if (math.className === 'MathJax_MathML') {
+        math = math.firstElementChild;
+      }
+      if (!math) return;
+      math.onkeydown = Explorer.Keydown;
```
dpvc: Should this use `addEventListener()` rather than setting `onkeydown` directly?

extensions/Assistive-Explore.js
```diff
@@ -195,46 +228,108 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+      //
+      Explorer.AddSpeech(math, script);
+      //
+      Explorer.Flame(math);
+      math.addEventListener(
```
dpvc: Will these listeners get added a second time if the same equation is walked twice?  I don't see anywhere where they are removed.

extensions/Assistive-Explore.js
```diff
@@ -331,21 +442,26 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
       'semantic': sre.SemanticWalker,
       'dummy': sre.DummyWalker
     },
-    ActivateWalker: function(math) {
+    ActivateWalker: function(math, jax) {
       var speechGenerator = new sre.DirectSpeechGenerator();
       var constructor = Explorer.Walkers[Assistive.getOption('walker')] ||
             Explorer.Walkers['dummy'];
-      Explorer.walker = new constructor(math, speechGenerator);
+      var mathml = jax.root.toMathML();
```
dpvc: This could potentially cause a file to be loaded (the circumstances are rare), and that is signaled by throwing an error.  You might at least want to use `try...catch` here to avoid failure in that situation.

extensions/Assistive-Explore.js
zorkow: Done.

extensions/Assistive-Explore.js
zorkow: Since you can only have one inline event at a time, it would simply replace the old listener (unlike addeventlistener, where they are added). But AddEvent is only ever called once on New Math, that should not make a difference.

extensions/Assistive-Explore.js
zorkow: I could, it would it make uniform with the other events that are added. At the time I did not want it to be interfered with by other keyboard events on the same node.
And I used addEventListener for the focus events primarily since they are not the same for all browsers. (Actually focusout is.)

extensions/Assistive-Explore.js
zorkow: I think I need more explanation for this:
How can a file be loaded on toMathML? I assume toMathML.js, but I would assume we must have loaded it already, since the extension pretty much depends on MathML (at least the enrichment and the walking).

extensions/Assistive-Explore.js
zorkow: Just remembered the reason: We need to overwrite, as with otherwise space will trigger the menu. So if we press "shift space" the walker will start and the menu will open.
If we overwrite we have to start with menu explicitly (which of course might not be loaded) as we currently do.
As an alternative, we could move to a different key, e.g., Enter. We had user requests for that already. Apparently users are more familiar with this from other solutions. @dpvc @pkra Any views on that?

dpvc: > We need to overwrite, as with otherwise space will trigger the menu

Ah, of course.

> As an alternative, we could move to a different key, e.g., Enter.

I'm fine with either way.  I guess it should be what people expect.  For the focused element (like a button) is it space or enter that triggers it (or both, or does it vary by browser)?

extensions/Assistive-Explore.js
dpvc: > But AddEvent is only ever called once on New Math

OK, got it.  Doesn't matter then, as you say.

extensions/Assistive-Explore.js
dpvc: >  I assume toMathML.js

No, that is not the file I mean.  The `toMathML()` method would not even be available if `toMathML.js` were not loaded.  I'm talking about situations where calling `toMathML()` can cause files to be loaded.  Does this extension *require* that `Semantic-MathML.js` also be included? I though that the syntactic walker didn't require that.  If the semantic enrichment is guaranteed to have run, then I withdraw my complaint here.

extensions/Assistive-Explore.js
zorkow: I've started a discussion on slack for that.
https://mathjax.slack.com/archives/a11y-tool/p1461790007000002

extensions/Assistive-Explore.js
zorkow: We always require semantic-mathml as we always need enriched MathML, even in the syntactic walker case, because, we effectively walk on the embedded semantic attributes.

dpvc: OK, then, this should be fine.  (The error handle may need to be done in Semantic-MathML in that case.)

Issue #98, opened 2016-05-03
Problem with annotation element.


zorkow:
If a MathML semantics annotation element (apparently Mathtype adds those) stops rendering of the enriched expression.
Simple example:

```
<math><semantics><mrow><mi>a</mi><mi>b</mi></mrow><annotation>implicit times</annotation></semantics></math>
```

zorkow:
@dpvc I've tested that a bit and it does not seem to have anything to do with the explorer. 
E.g., when you take out the assistive-explore extension entirely, it still does not render.
The enriched MathML looks correct, though.

dpvc:
It is in the Semantic-Collapse code.  I just pushed a fix for it in the `issue98` branch.  I'll make a PR.

PR #99, opened 2016-05-03
Don't descend into text nodes or nodes with no content.  Resolves issue #98

dpvc:
This fixes a problem with `<annotation>` elements throwing an error.  Resolves issue #98.


zorkow:
lgtm.

Issue #100, opened 2016-05-03
Complexity is set as undefined in the DOM elements


dpvc:
Something is wrong with the transferring of `complexity` value to the DOM elements created.

dpvc:
This is fixed in the `issue100` branch.

PR #101, opened 2016-05-04
Add complexity values to DOM nodes as data-semantic-complexity

dpvc:
This fixes the problem with the complexity values on the DOM nodes (issue #100), and make the attribute name `data-semantic-complexity` rather than `complexity` so that it is valid HTML.

### DIFF
--- a/extensions/Semantic-Complexity.js
+++ b/extensions/Semantic-Complexity.js
@@ -623,6 +623,13 @@ MathJax.Hub.Register.StartupHook("Semantic MathML Ready", function () {
       return this.complexity;
     }
   });
+  MML.annotation.Augment({
+    getComplexity: function () {
+      if (this.complexity == null) this.attrNames.push(COMPLEXATTR);
+      this.attr[COMPLEXATTR] = this.complexity = COMPLEXITY.XML;
+      return this.complexity;
+    }
+  });
 
   //
   //  Use fixed complexity, since we can't really measure it
--- a/extensions/Semantic-Complexity.js
+++ b/extensions/Semantic-Complexity.js
@@ -9,12 +9,14 @@
   var MML;
   
   var NOCOLLAPSE = 10000000; // really big complexity
+  var COMPLEXATTR = "data-semantic-complexity";
 
   var Complexity = MathJax.Extension.SemanticComplexity = {
     version: "1.0",
     config: HUB.CombineConfig("SemanticComplexity",{
       disabled: false
     }),
+    COMPLEXATTR: COMPLEXATTR,
 
     /*****************************************************************/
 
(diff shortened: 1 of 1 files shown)


extensions/Semantic-Complexity.js
```diff
@@ -526,8 +534,8 @@ MathJax.Hub.Register.StartupHook("Semantic MathML Ready", function () {
         if (this.data[this.sub])  C += COMPLEXITY.CHILD;
```
zorkow: Is there a reason why for underover we take the complexity of the base into account, but not for subsup?

zorkow:
Looks good. 
One question: You use == most of the time in the predicates. Since there is no need for type conversion, would === not be better? It will probably not matter very much for '== null', but generally it it is faster as there is no type conversion.

extensions/Semantic-Complexity.js
dpvc: Because the complexity is a surrogate for width, the layout of the expression has to be taken into account.

For msubsup, the larger of the complexities of the two scripts is added to the complexity of the base (analogous to the wider of the two scripts being added to the width of the base).

For munderover, on the other hand, the width would be the widest of the widths of the base, over and under elements.  That is taken care of in [lines 530-533](https://github.com/mathjax/MathJax-RespEq/blob/issue100/extensions/Semantic-Complexity.js#L533).  Note that the base *is* used here. 

In both cases, there is *extra* complexity added in for having children (and both take into account all three parts).

zorkow:
Just wrote this on git. Might fit with this PR:

Just writing some unit tests for semantics/annotation/annotation-xml elements.  Here's a pathological case:

<semantics><annotation>something</annotation></semantics>

I am not sure if that is legal. Spec says semantics "may" have mathml at first child, not must. It renders empty in MathJax but rendering breaks once enriched, probably again because complexity can't be computed.

@dpvc Do we need to be able to handle this?

dpvc:
> Since there is no need for type conversion, would `===` not be better? It will probably not matter very much for `== null`

I use `===` for all checks except this against `null`, because I want to check for both `null` and `undefined` at once.  Note that `null !== undefined` is true, but `null != undefined` is false.  Because elements in objects are sometimes non-existend (`undefined`), and sometimes cleared (set equal to `null`), I want a weaker check for `null`, in general.  It would probably be OK to use `===` in these cases, but I've gotten bitten by that with `null` in the past, and so have gotten in the habit of using `== null` rather than `=== null`.

zorkow:
Makes sense.

extensions/Semantic-Complexity.js
zorkow: I see. Thanks for the explanation.

zorkow:
I meant "wrote this on slack" in my earlier comment. Can't edit it anymore...

dpvc:
> Spec says semantics "may" have mathml at first child, not must.

I think that is a misreading.  It says the first child is the element to be annotated, so that can't itself be an annotation.  It says "Any MathML expression may appear as the first child", which doesn't say the first child may be MathML, it say the first child *is* MathML, and it may be *any* MathML element.

There is a use-case originating in Firefox that had an `<annotation-xml>` element as the first element, in which case that element was rendered.  This was the only way to include HTML in MathML and have the HTML displayed in firefox originally.  MathJax implements that as well, because Firefox did it, and it is still the only way to get HTML into MathJax output.  (HTML5 allows HTML interspersed with MathML, but we don't handle that.)  

Apparently this was an accidentally side effect of the way `<semantics>` was implemented in Firefox, and wa never intended by the spec.  This was discussed on the w3c math list at one point, if I recall properly, and that was made clear there.  So the example you give is invalid.

I think that the fix in #99 should resolve the error you are getting, however.  Are you using a current copy of master?  (It was only merged today.)

dpvc:
OK, I checked, and #99 doesn't fix annotation, just annotation-xml.  I can fix annotation, even though it is invalid MathML.  I suppose it would be best if we didn't crash.  :-)

zorkow:
Cleared cache and everything. Does not work.

zorkow:
This is interesting:
This works
<math><semantics><annotation><something/></annotation></semantics></math>
This does not
<math><semantics><annotation>something</annotation></semantics></math>

Anyway, it's all pathological cases.

dpvc:
I've added the code to handle the `<annotation>` element.  No big deal.

zorkow:
Works!

PR #109, opened 2016-05-11
Lightweight extension

zorkow:
All one needs to load now is the Accessibility-Extension.js file.
This inserts three new menu items (Responsive equations, Auto Collapse, Accessibility) into the context menu, that allow to load the other modules selectively.
Directly including those modules still works as well.
Accessibility menu can be switched off again with a checkbox, which I called Active for now. Not sure if that is a good name. The other two are just checkboxes but could easily be expanded.

(There is still an issue when switching off the speech, that the message hook does not go away. See https://github.com/mathjax/MathJax-RespEq/compare/abstraction_rules...lightweight_extension?expand=1#diff-3d0281e358905009434a406ce04579d4R101
)


dpvc:
I have been looking at this for the last few days, and have found a few issues (mostly timing problems).  I'm working on a pull request to resolve them, but some of it involves some reorganizing of pieces of it to allow the timing to work out properly, so it is taking longer than I had hoped.  I should have something soon, and will let you know.  Sorry for the delay!

Issue #105, opened 2016-05-07
Complexity


zorkow:
`data-semantic-complexity` can appear twice in MathML element. That leads to an XML error at my end.
I believe it only happens for mactions on the top level.
Example expression:

i = 1, 2, \ldots, n+1

dpvc:
I'm not able to reproduce this on the master branch.  Is it one of the other branches that causes this?  If not, is there something more that I'm supposed to do other than just enter the TeX expression above into the Semantics TeX lab?

zorkow:
This is what I get on master branch:
<mrow complexity="24" speech="i equals one comma two comma ellipsis comma n plus one" content="3,5,7" children="11,3,4,5,6,7,12" id="13" role="sequence" type="punctuated" complexity="24">

Uses 
MathJax.Ajax.config.path["SRE"] = "https://mathjax.github.io/MathJax-accessibility/dist";
and the above expression.

dpvc:
OK, I see it now.  Thanks!

dpvc:
I have submitted a pull request that fixes this issue.

pkra:
Fixed by #110.

PR #110, opened 2016-05-17
Remove duplicate complexity added maction's mrow.  Resolves issue #105.

dpvc:
Handle adding of complexity attribute better, and remove duplicate for `<mrow>` in complexity generated `<maction>`.  Resolves issue #105.

This now uses a new `reportComplexity()` method that checks to see if the complexity has already been added.  The main culprit for the duplicate was lines 183 and 184 (in conjunction with the loop right before that, which already added the `data-semantic-complexity` attribute).

### DIFF
--- a/extensions/Semantic-Complexity.js
+++ b/extensions/Semantic-Complexity.js
@@ -180,8 +180,7 @@
             mml.attrNames.splice(i,1);
           }
         }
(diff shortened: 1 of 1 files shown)


zorkow:
Thanks. Works.

PR #112, opened 2016-05-23
Patch to Volker's extension to fix timing issues, etc.

dpvc:
Patch to Volker's extension to fix timing issues, prevent extra reprocess, handle configuration and menu cookies better, etc.

I will write more about this tomorrow.  I'm going to be unavailable the rest of the day, but wanted to get this started so you could look at it before Wednesday, if you wanted.  Some restructuring was needed in order to make some parts of the timing work better.  As I say, I'll explain it further soon.

### DIFF
--- a/extensions/Accessibility-Extension.js
+++ b/extensions/Accessibility-Extension.js
@@ -12,6 +12,15 @@
     for (var id in obj) {if (obj.hasOwnProperty(id)) keys.push(id)}
     return keys;
   };
+  
+  //
+  //  Set up the a11y path,if it isn't already in place
+  //
+  var PATH = MathJax.Ajax.config.path;
+  if (!PATH.a11y) PATH.a11y =
+      (PATH.Contrib ? PATH.Contrib + "/a11y" : 
+      (String(location.protocal).match(/^https?:/) ? "" : "http:") + 
+        "//cdn.mathjax.org/mathjax/contrib/a11y");
 
   var Accessibility = EXTENSIONS.Accessibility = {
     version: '1.0',
@@ -117,19 +126,19 @@
   
   Accessibility.Register(
     ModuleLoader(
-      'collapsible', 'Collapsible Math', '[RespEq]/Semantic-Complexity.js',
+      'collapsible', 'Collapsible Math', '[a11y]/Semantic-Complexity.js',
       'SemanticComplexity'
     )
   );
   Accessibility.Register(
     ModuleLoader(
-      'autocollapse', 'Auto Collapse', '[RespEq]/Semantic-Collapse.js',
+      'autocollapse', 'Auto Collapse', '[a11y]/Semantic-Collapse.js',
       'SemanticCollapse'
     )
   );
   Accessibility.Register(
     ModuleLoader(
-      'explorer', 'Accessibility', '[RespEq]/Assistive-Explore.js',
+      'explorer', 'Accessibility', '[a11y]/Assistive-Explore.js',
       'Assistive', true
     )
   );
@@ -138,7 +147,7 @@
   
   MathJax.Callback.Queue(
     ["LoadExtensions",Accessibility],
-    ["loadComplete",MathJax.Ajax,"[RespEq]/Accessibility-Extension.js"]
+    ["loadComplete",MathJax.Ajax,"[a11y]/Accessibility-Extension.js"]
   );
 
 })(MathJax.Hub,MathJax.Extension);
--- a/extensions/Accessibility-Extension.js
+++ b/extensions/Accessibility-Extension.js
@@ -6,7 +6,7 @@
   var ITEM, MENU; // filled in when MathMenu extension loads
   
   var BIND = (Function.prototype.bind ? function (f,t) {return f.bind(t)} :
-              function (f,t) {return function () f.apply(t,arguments)});
+              function (f,t) {return function () {f.apply(t,arguments)}});
   var KEYS = Object.keys || function (obj) {
     var keys = [];
     for (var id in obj) {if (obj.hasOwnProperty(id)) keys.push(id)}
@@ -53,7 +53,7 @@
     },
     Startup: function() {
       ITEM = MathJax.Menu.ITEM;
-      MENU = MathJax.Menu.placeHolder;
+      MENU = MathJax.Menu.menu;
       for (var i = 0, module; module = this.modules[i]; i++) {
         module.CreateMenu();
         module.AddMethods();
@@ -117,11 +117,12 @@
     }
   });
 
-  HUB.Register.StartupHook('MathMenu Ready', function() {
-    Accessibility.Startup();
-    HUB.Startup.signal.Post('Accessibility Loader Ready');
-  },5);   // make sure it runs before other extensions' menu hooks
-          //   even if they are loaded first
+  HUB.Register.StartupHook('End Extensions', function () {
+    HUB.Register.StartupHook('MathMenu Ready', function () {
+      Accessibility.Startup();
+      HUB.Startup.signal.Post('Accessibility Loader Ready');
+    },5);   // run before other extensions' menu hooks even if they are loaded first
+  },5);
   
   Accessibility.Register(
     ModuleLoader(
--- a/extensions/Accessibility-Extension.js
+++ b/extensions/Accessibility-Extension.js
@@ -1,12 +1,19 @@
 //
 // Thin hook to include the accessibility extension.
 //
-(function(HUB) {
+(function(HUB,EXTENSIONS) {
   var SETTINGS = HUB.config.menuSettings;
-  var ITEM = MathJax.Menu.ITEM;
-  var MENU = MathJax.Menu.menu;
+  var ITEM, MENU; // filled in when MathMenu extension loads
+  
+  var BIND = (Function.prototype.bind ? function (f,t) {return f.bind(t)} :
+              function (f,t) {return function () f.apply(t,arguments)});
+  var KEYS = Object.keys || function (obj) {
+    var keys = [];
+    for (var id in obj) {if (obj.hasOwnProperty(id)) keys.push(id)}
+    return keys;
+  };
 
-  var Accessibility = MathJax.Extension.Accessibility = {
+  var Accessibility = EXTENSIONS.Accessibility = {
     version: '1.0',
     prefix: '', //'Accessibility-',
     default: {},
@@ -18,7 +25,7 @@
       return SETTINGS[Accessibility.MakeOption(option)];
     },
     AddDefaults: function() {
-      var keys = Object.keys(Accessibility.default);
+      var keys = KEYS(Accessibility.default);
       for (var i = 0, key; key = keys[i]; i++) {
         var option = Accessibility.MakeOption(key);
         if (typeof(SETTINGS[option]) === 'undefined') {
@@ -28,7 +35,8 @@
     },
     // Attaches the menu items;
     AddMenu: function() {
-      var items = Accessibility.modules.map(function(x) {return x.placeHolder;});
+      var items = Array(this.modules.length);
+      for (var i = 0, module; module = this.modules[i]; i++) items[i] = module.placeHolder;
       var about = MENU.IndexOfId('About');
       if (about === null) {
         items.unshift(ITEM.RULE());
@@ -44,124 +52,103 @@
       Accessibility.modules.push(module);
     },
     Startup: function() {
-      Accessibility.AddDefaults();
-      Accessibility.AddMenu();
-      var result = Accessibility.modules.map(function(x) {
-        return x.Load();});
-      if (result.reduce(function(x, y) {return x || y;}, false)) {
-        HUB.Queue(['Reprocess', HUB]);
+      ITEM = MathJax.Menu.ITEM;
+      MENU = MathJax.Menu.placeHolder;
+      for (var i = 0, module; module = this.modules[i]; i++) {
+        module.CreateMenu();
+        module.AddMethods();
       }
+      this.AddMenu();
+    },
+    LoadExtensions: function () {
+      var extensions = [];
+      for (var i = 0, mpdule; module = this.modules[i]; i++) {
+        if (SETTINGS[module.option]) extensions.push(module.module);
+      }
+      return (extensions.length ? HUB.Startup.loadArray(extensions) : null);
     }
   };
-
 
   var ModuleLoader = MathJax.Extension.ModuleLoader = MathJax.Object.Subclass({
     option: '',
     name: '',
     module: '',
-    signal: '',
     placeHolder: null,
-    activeBox: null,
-    subMenu: null,
+    submenu: false,
     extension: null,
     Enable: null,
     Disable: null,
-    onSwitch: false,
-    Init: function(option, name, module, signal, extension) {
+    Init: function(option, name, module, extension, submenu) {
       this.option = option;
-      this.name = name;
+      this.name = [name.replace(/ /g,''),name];
       this.module = module;
-      this.signal = signal;
       this.extension = extension;
-      var that = this;
-      var helper = function() {that.Switch.apply(that);};
-      this.placeHolder = ITEM.CHECKBOX(
-        this.name, Accessibility.MakeOption(this.option), {action: helper});
-      this.activeBox = ITEM.CHECKBOX(
-        'Active', Accessibility.MakeOption(this.option), {action: helper});
+      this.submenu = (submenu || false);
+    },
+    CreateMenu: function() {
+      var load = BIND(this.Load,this);
+      if (this.submenu) {
+        this.placeHolder =
+          ITEM.SUBMENU(this.name,
+            ITEM.CHECKBOX(["Activate","Activate"],
+                          Accessibility.MakeOption(this.option), {action: load}),
+            ITEM.RULE(),
+            ITEM.COMMAND(["OptionsWhenActive","(Options when Active)"],null,{disabled:true})
+          );
+      } else {
+        this.placeHolder = ITEM.CHECKBOX(
+          this.name, Accessibility.MakeOption(this.option), {action: load}
+        );
+      }
     },
     Load: function() {
-      var flag = Accessibility.GetOption(this.option);
-      if (flag) {
-        if (this.subMenu) {
-          this.AddSubMenu();
-        } else {
-          MathJax.Ajax.Require(this.module);
-          this.AddMethods();
-          this.AddActiveBox();
+      HUB.Queue(["Require",MathJax.Ajax,this.module,["AddMethods",this,true]]);
+    },
+    AddMethods: function(menu) {
+      var ext = MathJax.Extension[this.extension];
+      if (ext) {
+        this.Enable = BIND(ext.Enable,ext);
+        this.Disable = BIND(ext.Disable,ext);
+        if (menu) {
+          this.Enable(true,true);
+          MathJax.Menu.saveCookie();
         }
-      }
-      return flag;
-    },
-    Switch: function() {
-      this.onSwitch = true;
-      var flag = this.Load();
-      if (!flag) {
-        this.AddPlaceHolder();
-        this.Disable(true);
-      } else {
-        if (this.subMenu) {
-          this.Enable(true);
-        }
-      }
-    },
-    AddMethods: function() {
-      var that = this;
-      HUB.Register.StartupHook(this.signal, function() {
-        var ext = MathJax.Extension[that.extension];
-        if (ext) {
-          that.Enable = ext.Enable;
-          that.Disable = ext.Disable;
-          if (that.onSwitch) that.Enable(true);
-        }
-      });
-    },
-    AddActiveBox: function() {
-      var that = this;
-      HUB.Register.StartupHook(this.signal, function() {
-        if (that.subMenu) return;
-        var submenu = MENU.FindId(that.name);
-        if (submenu !== null && submenu.submenu) {
-          submenu.submenu.items.push(ITEM.RULE(), that.activeBox);
-        }
-      });
-    },
-    AddPlaceHolder: function() {
-      var index = MENU.IndexOfId(this.name);
-      if (index !== null) {
-        this.subMenu = MENU.items[index];
-        MENU.items[index] = this.placeHolder;
-      }
-    },
-    AddSubMenu: function() {
-      var index = MENU.IndexOfId(this.name);
-      if (index !== null) {
-        MENU.items[index] = this.subMenu;
       }
     }
-  },{});
+  });
 
-})(MathJax.Hub);
+  HUB.Register.StartupHook('MathMenu Ready', function() {
+    Accessibility.Startup();
+    HUB.Startup.signal.Post('Accessibility Loader Ready');
+  },5);   // make sure it runs before other extensions' menu hooks
+          //   even if they are loaded first
+  
+  Accessibility.Register(
+    ModuleLoader(
+      'collapsible', 'Collapsible Math', '[RespEq]/Semantic-Complexity.js',
+      'SemanticComplexity'
+    )
+  );
+  Accessibility.Register(
+    ModuleLoader(
+      'autocollapse', 'Auto Collapse', '[RespEq]/Semantic-Collapse.js',
+      'SemanticCollapse'
+    )
+  );
+  Accessibility.Register(
+    ModuleLoader(
+      'explorer', 'Accessibility', '[RespEq]/Assistive-Explore.js',
+      'Assistive', true
+    )
+  );
 
+  Accessibility.AddDefaults();
+  
+  MathJax.Callback.Queue(
+    ["LoadExtensions",Accessibility],
+    ["loadComplete",MathJax.Ajax,"[RespEq]/Accessibility-Extension.js"]
+  );
 
+})(MathJax.Hub,MathJax.Extension);
 
-MathJax.Callback.Queue(
-  MathJax.Hub.Register.StartupHook('MathMenu Ready', function() {
-  MathJax.Extension.Accessibility.Register(
-    MathJax.Extension.ModuleLoader(
-      'responsive', 'Responsive Equations', '[RespEq]/Semantic-Complexity.js',
-      'Semantic Complexity Ready', 'SemanticComplexity'));
-    MathJax.Extension.Accessibility.Register(
-      MathJax.Extension.ModuleLoader(
-        'collapse', 'Auto Collapse', '[RespEq]/Semantic-Collapse.js',
-        'Semantic Collapse Ready', 'SemanticCollapse'));
-    MathJax.Extension.Accessibility.Register(
-      MathJax.Extension.ModuleLoader(
-        'explorer', 'Accessibility', '[RespEq]/Assistive-Explore.js',
-        'Explorer Ready', 'Assistive'));
-    MathJax.Extension.Accessibility.Startup();
-  MathJax.Hub.Startup.signal.Post('Accessibility Loader Ready');
-}));
-
-MathJax.Ajax.loadComplete("[RespEq]/Accessibility-Extension.js");
 
--- a/extensions/Accessibility-Extension.js
+++ b/extensions/Accessibility-Extension.js
@@ -54,10 +54,7 @@
     Startup: function() {
       ITEM = MathJax.Menu.ITEM;
       MENU = MathJax.Menu.menu;
-      for (var i = 0, module; module = this.modules[i]; i++) {
-        module.CreateMenu();
-        module.AddMethods();
-      }
+      for (var i = 0, module; module = this.modules[i]; i++) module.CreateMenu();
       this.AddMenu();
     },
     LoadExtensions: function () {
@@ -71,13 +68,11 @@
 
   var ModuleLoader = MathJax.Extension.ModuleLoader = MathJax.Object.Subclass({
     option: '',
-    name: '',
+    name: ['',''],
     module: '',
     placeHolder: null,
     submenu: false,
     extension: null,
-    Enable: null,
-    Disable: null,
     Init: function(option, name, module, extension, submenu) {
       this.option = option;
       this.name = [name.replace(/ /g,''),name];
@@ -102,17 +97,13 @@
       }
     },
     Load: function() {
-      HUB.Queue(["Require",MathJax.Ajax,this.module,["AddMethods",this,true]]);
+      HUB.Queue(["Require",MathJax.Ajax,this.module,["Enable",this]]);
     },
-    AddMethods: function(menu) {
-      var ext = MathJax.Extension[this.extension];
-      if (ext) {
-        this.Enable = BIND(ext.Enable,ext);
-        this.Disable = BIND(ext.Disable,ext);
-        if (menu) {
-          this.Enable(true,true);
-          MathJax.Menu.saveCookie();
-        }
+    Enable: function(menu) {
+      var extension = MathJax.Extension[this.extension];
+      if (extension) {
+        extension.Enable(true,true);
+        MathJax.Menu.saveCookie();
       }
     }
   });
--- a/extensions/Assistive-Explore.js
+++ b/extensions/Assistive-Explore.js
@@ -128,6 +128,12 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
           for (var i = 2, item; item = items[i]; i++) item.disabled = state;
         }
       }
+    },
+    //
+    //  Register a dependant
+    //
+    Dependant: function (extension) {
+      this.dependants.push(extension);
     }
   };
 
@@ -640,7 +646,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
     },
     Startup: function() {
       var Complexity = MathJax.Extension.SemanticComplexity;
-      if (Complexity) Complexity.dependants.push(Assistive);
+      if (Complexity) Complexity.Dependant(Assistive);
       Assistive.addDefaults();
       this.createMenu();
     },
--- a/extensions/Assistive-Explore.js
+++ b/extensions/Assistive-Explore.js
@@ -733,7 +733,6 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
             index = MENU.IndexOfId('CollapsibleMath');
             MENU.items.splice(index+1,0,accessibilityMenu);
           }
-          Assistive[SETTINGS.explorer ? "Enable" : "Disable"]();
         },20);  // Between Semantic-Complexity and Semantic-Collapse
       },20);
     }
--- a/extensions/Assistive-Explore.js
+++ b/extensions/Assistive-Explore.js
@@ -436,7 +436,6 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
     // Event execution on keydown. Subsumes the same method of MathEvents.
     //
     Keydown: function(event) {
-      if (!Assistive.hook) return;
       if (event.keyCode === KEY.ESCAPE) {
         if (!Explorer.walker) return;
         Explorer.RemoveHook();
@@ -473,7 +472,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
       }
       var math = event.target;
       if (event.keyCode === KEY.SPACE) {
-        if (event.shiftKey) {
+        if (event.shiftKey && Assistive.hook) {
           var jax = MathJax.Hub.getJaxFor(math);
           Explorer.ActivateWalker(math, jax);
           Explorer.AddHook(jax);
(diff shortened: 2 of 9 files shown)


zorkow:
Thanks. I get a couple of errors. I'll try to put some details into inline comments.

PS: I've started rewriting the Menu as discussed last week, but will hold off until this part is merged.

extensions/Accessibility-Extension.js
```diff
@@ -1,12 +1,19 @@
...
 //
-(function(HUB) {
+(function(HUB,EXTENSIONS) {
   var SETTINGS = HUB.config.menuSettings;
-  var ITEM = MathJax.Menu.ITEM;
-  var MENU = MathJax.Menu.menu;
+  var ITEM, MENU; // filled in when MathMenu extension loads
+  
+  var BIND = (Function.prototype.bind ? function (f,t) {return f.bind(t)} :
+              function (f,t) {return function () f.apply(t,arguments)});
```
zorkow: Arguments is not defined. That leads to a bug in Chrome (not FF).

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,103 @@
...
     },
     Startup: function() {
-      Accessibility.AddDefaults();
-      Accessibility.AddMenu();
-      var result = Accessibility.modules.map(function(x) {
-        return x.Load();});
-      if (result.reduce(function(x, y) {return x || y;}, false)) {
-        HUB.Queue(['Reprocess', HUB]);
+      ITEM = MathJax.Menu.ITEM;
+      MENU = MathJax.Menu.placeHolder;
```
zorkow: This is undefined. Placeholders are instantiated after the CreateMenu method in the subsequent loop.
However, I assume that should be 
MENU = MathJax.Menu.menu 
anyway.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,103 @@
...
-
-})(MathJax.Hub);
+  });
 
+  HUB.Register.StartupHook('MathMenu Ready', function() {
+    Accessibility.Startup();
+    HUB.Startup.signal.Post('Accessibility Loader Ready');
+  },5);   // make sure it runs before other extensions' menu hooks
+          //   even if they are loaded first
+
```
zorkow: This should go after line 145, as o/w the modules might not exist in the AddMenu method.

zorkow:
Now that I got it to work, I see that you have started moving the 
So I'll stop with my rewrite entirely. Looks great with the Active at the top and the grayed out options when you switch it off.
However, when I switch off accessibility, I also can't get into the context menu anymore with Space.

extensions/Accessibility-Extension.js
dpvc: OOPS, I was doing a global find-and-replace at the last minute and that was accidentally changed.  (I had changed `placeHolder` to `menu` at one point, and then decided to put it back.)

extensions/Accessibility-Extension.js
dpvc: arguments should *always* be defined within a function.  It refers to the arguments of the function.  Did I misspell it?   In any case, Chrome should never execute that statement, since it does have the `bind` function.

extensions/Accessibility-Extension.js
dpvc: The `MathMenu Ready` signal should not occur until after the extensions are loaded (but perhaps that isn't true with a combined configuration file).  But if that is the case, after 145 isn't right either.  I will check into it.

dpvc:
> when I switch off accessibility, I also can't get into the context menu anymore with Space

That is probably because of the check I added to the `Keypress` method.  That will need to be made more sophisticated.  Sorry about that.

dpvc:
I added some replies to your comments, but I don't see them above.  Did you get copies, or should I add them again?

zorkow:
Same here. I've been frantically trying to reload the PR but they never
appeared.
Will reply by email.


On Mon, May 23, 2016 at 4:32 PM, Davide P. Cervone <notifications@github.com

extensions/Accessibility-Extension.js
zorkow: Forgot about arguments...
Anyway, tt's a scoping issue. Butting  { } around  the body of the second
function will get rid of the problem.
Why Chrome evaluates this part, I really don't know.

(I always like { } around bodies... not only because my linter tells me
to...)

dpvc: D'oh.  They certainly should have been there.  I was using FF to debug, so didn't get any message.  Apologies.

dpvc:
It looks like the line comments made it in after all (though I guess it took a while to register).  I have fixed the issues you identified, and have push some more commits to take care of them.  I think it is ready now.  I will outline the major changes below, and make line comments explaining some of the details in line comments.

As you know, the synchronization of timing between the extensions, and with the MathMenu extension is complicated but critical.  So a number of changes are there to coordinate that better.  You pointed out that even what I had wasn't good enough (when you suggested moving of the StartupHook).  The MathMenu extension can be loaded in a variety of ways that complicate the timing further:  it can be in a combined configuration file, which will be loaded before the extensions (and the `MathMenu Ready` will have already fired when the extension are loaded, so their MathMenu hooks will fire in the order that they are loaded, which is random), or the MathMenu could be loaded in the `extensions` array along with other extensions (rare, but allowed), in which case the `MathMenu Ready` signal will fire at a random time among the extensions (some might be laded already and others not), or MathMenu could be loaded automatically by MathJax several seconds after the initial typeset (if you don't use a combined configuration file and don't include it in the extensions explicitly).  So the point at which the menu hooks are run can vary greatly, and since the order in which the various extensions have their menus processed is important, we need to take all those possibilities into account.

I believe the code now does that.  The key is to use nested startup hooks:  one that waits for the `End Extensions` signal which will always happen after all the extensions are loaded (other than possibly the MathMenu extension, which can be loaded automatically later), and then a second one for the `MathMenu Ready` signal.  For both signals, we use handler priorities to force the order to be the one we want, regardless of the order in which extensions are loaded, and that means that the `MathMenu Ready` signals will run in the correct order, regardless of when MathMenu is loaded in relation to the extensions.  (If MathMenu comes first, we still get the proper order from the `End Extensions` order, and if it comes later, we get the correct order due to the priorities.)

The Assistive-Extension's MathMenu hook needs to run first, so that it creates the menus before the other extensions try to do so.  When the individual extensions are loaded, they look for the menu created by Assistive-Extension, and replace it if found, otherwise they add their own menu.  In the latter case, rather than always adding a rule and their own menu, by forcing the order in which the menu handlers are fired, we can replicate the menu setup used by Assistive-Extension, even if it isn't loaded.  

If the Assistive-Extension doesn't load, the Semantic-Complexity's MathMenu hook runs first and adds the first menu (and a rule).  Assistive-Explorer (which forces Semantic-MathML to load) runs next, and looks for the menu from Semantic-Complexity, which it knows will be there, and adds the Assistive menu below that.  Then the Semantic-Collapse hook runs, and again, since it forces Semantic-Complexity, it knows its menu will be there, and adds its menu in after that (which means it is between "Collapsible Math" and "Assistive", where it belongs.

Note that these can be loaded in any order, and any combination, and the result should be correctly ordered (for those menus that appear), and that this works whether Assistive-Extension is loaded or not.  (It took some effort to make that all work out.)

Since the extensions can be loaded without the Assistive-Extension, they have to be able to manage their own menus (i.e., enable and disable themselves).  Once extension is loaded, there is no real need for the Assistive-Extension to be involved with that menu any longer; it is really only there to manage the place-holder menu that will force the loading of an extension that was not active originally but is activated by the user.  So the Assistive-Extension ModuleLoader has been simplified to only deal with loading the extensions, not managing the menu after that.  Once loaded, the extension will replace the place-holder menu and it manage enable/disable actions itself (as they would if they were loaded but that page in the absence of the Assistive-Extension).

This also means that the Enable and Disable operations need to be a bit more sophisticated, since they need to make sure that the required extensions are also enabled, or dependent extensions are disabled.  So these methods have been expanded to properly signal the other extensions.  They also have to make sure the menu cookie is updated (the menu code handles updating the cookie for the extension that is endabled/disabled, but if that forces others to be enabled/disabled, those changes need to be recorded in the cookie as well).

Another timing issue is that e would like to include whatever enrichment is to be done to the math on the *first* typesetting that occurs, rather than do one typesetting without enrichment and then reprocess it again for enrichment.  This is what was happening earlier, but that leads to a poor user experience since they have to wait for the page to be typeset twice.

In order to avoid the double typeset, we need to make sure that any of the active extensions are loaded before the initial typesetting occurs.  The solution is to force the extensions to load *before* signaling that the Assistive-Extension is loaded, so that MathJax will wait for *all* the needed extensions before doing its initial typesetting pass.  That required refactoring the ModuleLoader code somewhat.

Finally, I changed the name of the first menu from "Responsive Equations" to "Collapsible Math", which is (a) shorter, and (b) more in line with it actual function: it makes the math collapsible, but it is not really "responsive" to anything ("Auto Collapse" is more the "responsive" option).

I'll make more comments in the code.

extensions/Accessibility-Extension.js
```diff
@@ -1,12 +1,19 @@
 //
 // Thin hook to include the accessibility extension.
 //
-(function(HUB) {
+(function(HUB,EXTENSIONS) {
   var SETTINGS = HUB.config.menuSettings;
-  var ITEM = MathJax.Menu.ITEM;
-  var MENU = MathJax.Menu.menu;
+  var ITEM, MENU; // filled in when MathMenu extension loads
```
dpvc: When not using a combined configuration file, the `MathJax.Menu` might not be defined yet (usually it is loaded after the initial typesetting).  So these need to be set by the MathMenu hook later.

extensions/Accessibility-Extension.js
```diff
@@ -1,12 +1,19 @@
...
+  
+  var BIND = (Function.prototype.bind ? function (f,t) {return f.bind(t)} :
+              function (f,t) {return function () {f.apply(t,arguments)}});
+  var KEYS = Object.keys || function (obj) {
+    var keys = [];
+    for (var id in obj) {if (obj.hasOwnProperty(id)) keys.push(id)}
+    return keys;
+  };
```
dpvc: These allow the code to run even in IE < 9, even though selecting the accessibility options will not actually work.  We might want to do something to disable the menus entirely for early versions of IE.

extensions/Accessibility-Extension.js
```diff
@@ -28,7 +35,8 @@
     },
     // Attaches the menu items;
     AddMenu: function() {
-      var items = Accessibility.modules.map(function(x) {return x.placeHolder;});
+      var items = Array(this.modules.length);
+      for (var i = 0, module; module = this.modules[i]; i++) items[i] = module.placeHolder;
       var about = MENU.IndexOfId('About');
```
dpvc: This is also for IE < 9

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-      if (result.reduce(function(x, y) {return x || y;}, false)) {
-        HUB.Queue(['Reprocess', HUB]);
+      ITEM = MathJax.Menu.ITEM;
+      MENU = MathJax.Menu.menu;
+      for (var i = 0, module; module = this.modules[i]; i++) {
+        module.CreateMenu();
+        module.AddMethods();
+      }
+      this.AddMenu();
+    },
```
dpvc: The loading of the extensions has to be done earlier, so we remove that from `Startup()`.  Since `Startup()` occurs in the MathMenu hook, we can set the `ITEM` and `MENU` values needed elsewhere.  We also create the place-holder menus here (since we had to create and register the `ModuleLoader`s before the MathMenu code may be available, and couldn't do it then).  We also add the methods (for any loaded modules).  Technically, this really isn't needed, since the menu now only loads the extension, and the extension will replace the menu, so the `ModuleLoader`s `Enable()` or `Disable()` method will not be called.  I may simplify this further.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
+    LoadExtensions: function () {
+      var extensions = [];
+      for (var i = 0, mpdule; module = this.modules[i]; i++) {
+        if (SETTINGS[module.option]) extensions.push(module.module);
       }
+      return (extensions.length ? HUB.Startup.loadArray(extensions) : null);
     }
   };
```
dpvc: This is used to load any extensions that are requested by the menu settings (i.e., ones that were checked by the user earlier or by the default configuration set up by the author).  This has to be done before the Assistive-Extension calls `loadComplete` so that we are sure all the needed extensions have been loaded before the initial typeset pass.  (See the call to `LoadExtensions` at the end of the file.)

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-
   var ModuleLoader = MathJax.Extension.ModuleLoader = MathJax.Object.Subclass({
     option: '',
     name: '',
     module: '',
-    signal: '',
     placeHolder: null,
-    activeBox: null,
-    subMenu: null,
+    submenu: false,
```
dpvc: This is now a signal for whether the extension's version of the menu will be a submenu (like `Assistive`) and should have an `Activate` entry, as we discussed in the hangout last week.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
     extension: null,
     Enable: null,
     Disable: null,
-    onSwitch: false,
-    Init: function(option, name, module, signal, extension) {
+    Init: function(option, name, module, extension, submenu) {
       this.option = option;
-      this.name = name;
+      this.name = [name.replace(/ /g,''),name];
```
dpvc: In order to be properly localized, we need to use `[id,name]` for the menu name, where the `id` is used for identifying the menu, and `name` is the string that will be localized (and changes from language to language).  Currently, the `id` is just the `name` with spaces removed, but in the future, the pair could be passed as the `name` argument directly.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
       this.module = module;
-      this.signal = signal;
       this.extension = extension;
-      var that = this;
-      var helper = function() {that.Switch.apply(that);};
-      this.placeHolder = ITEM.CHECKBOX(
-        this.name, Accessibility.MakeOption(this.option), {action: helper});
-      this.activeBox = ITEM.CHECKBOX(
-        'Active', Accessibility.MakeOption(this.option), {action: helper});
```
dpvc: We can't create the menu at this point because the `ModuleLoader` may be created before the MathMenu extension is loaded.  So we put the creation of place-holder menus into the `Startup()` method, which happens when MathMenu loads.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-        } else {
-          MathJax.Ajax.Require(this.module);
-          this.AddMethods();
-          this.AddActiveBox();
-        }
-      }
-      return flag;
+      this.submenu = (submenu || false);
     },
-    Switch: function() {
```
dpvc: We don't need a `Switch()` method any longer, since the only action the place-holder menu does is load the extension.  After that, the extension replaces the place-holder, and the place-holder is no longer used.  We never go back.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
+          );
       } else {
-        if (this.subMenu) {
-          this.Enable(true);
-        }
+        this.placeHolder = ITEM.CHECKBOX(
+          this.name, Accessibility.MakeOption(this.option), {action: load}
+        );
       }
     },
```
dpvc: The `CreateMenu()` method is called during `Startup()` to create place-holder menus that load the extensions.  The proper type of menu (`CHECKBOX` or `SUBMENU`) is created based on the needs of the `ModuleLoader`.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-    AddMethods: function() {
-      var that = this;
-      HUB.Register.StartupHook(this.signal, function() {
-        var ext = MathJax.Extension[that.extension];
-        if (ext) {
-          that.Enable = ext.Enable;
-          that.Disable = ext.Disable;
-          if (that.onSwitch) that.Enable(true);
-        }
-      });
```
dpvc: The modified version of this is below (I guess I inserted routines above this, and the `git diff` didn't get the connection between the two versions).

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
+    Load: function() {
+      HUB.Queue(["Require",MathJax.Ajax,this.module,["AddMethods",this,true]]);
     },
```
dpvc: The load action is to use `MathJax.Ajax.Require()` to load the function, calling `this.AddMethods(true)` when that is done.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-    AddActiveBox: function() {
```
dpvc: This is no longer needed.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-          submenu.submenu.items.push(ITEM.RULE(), that.activeBox);
+    AddMethods: function(menu) {
+      var ext = MathJax.Extension[this.extension];
+      if (ext) {
+        this.Enable = BIND(ext.Enable,ext);
+        this.Disable = BIND(ext.Disable,ext);
+        if (menu) {
+          this.Enable(true,true);
+          MathJax.Menu.saveCookie();
         }
```
dpvc: Here we get the Enable/Disable methods, and if called with `true` (i.e., from the load action of a place-holder menu), we enable the extension and save the menu cookie (since the enabling may have enabled other extensions as well whose cookie state needs to be saved).

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-        this.subMenu = MENU.items[index];
-        MENU.items[index] = this.placeHolder;
-      }
-    },
-    AddSubMenu: function() {
-      var index = MENU.IndexOfId(this.name);
-      if (index !== null) {
-        MENU.items[index] = this.subMenu;
       }
     }
```
dpvc: We don't need `AddPlaceHolder()` or `AddSubMenu()` any longer, as the extensions manage their own menus (and rep ace the place-holders) when they are loaded.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
+      'autocollapse', 'Auto Collapse', '[RespEq]/Semantic-Collapse.js',
+      'SemanticCollapse'
+    )
+  );
+  Accessibility.Register(
+    ModuleLoader(
+      'explorer', 'Accessibility', '[RespEq]/Assistive-Explore.js',
+      'Assistive', true
+    )
+  );
```
dpvc: We create and register the loaders immediately, since we need to know what extensions to load before `loadComplete()` is called (so they are loaded before the first typeset pass, avoiding an extra reprocess step and user frustration).

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
 
+  Accessibility.AddDefaults();
```
dpvc: We can now set the `menuSettings` for the registered extensions.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
+  
+  MathJax.Callback.Queue(
+    ["LoadExtensions",Accessibility],
+    ["loadComplete",MathJax.Ajax,"[RespEq]/Accessibility-Extension.js"]
+  );
```
dpvc: We load any extensions that have been selected by the user or the page author, and then signal that our loading is complete (so the MathJax startup process can continue on; the `End Extensions` signal should follow this).

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-      MathJax.Extension.ModuleLoader(
-        'collapse', 'Auto Collapse', '[RespEq]/Semantic-Collapse.js',
-        'Semantic Collapse Ready', 'SemanticCollapse'));
-    MathJax.Extension.Accessibility.Register(
-      MathJax.Extension.ModuleLoader(
-        'explorer', 'Accessibility', '[RespEq]/Assistive-Explore.js',
-        'Explorer Ready', 'Assistive'));
-    MathJax.Extension.Accessibility.Startup();
-  MathJax.Hub.Startup.signal.Post('Accessibility Loader Ready');
-}));
```
dpvc: All of this has to be done earlier, since this queued action occurs *after* the initial typeset pass.  Since we need to load extensions *before* that typesetting, the registration of modules needs to be done earlier, as discussed above, and in the main comments.

extensions/Assistive-Explore.js
```diff
@@ -4,6 +4,7 @@
 MathJax.Hub.Register.StartupHook('Sre Ready', function() {
   var FALSE, KEY;
   var SETTINGS = MathJax.Hub.config.menuSettings;
+  var COOKIE = {};   // replaced when menu is available
```
dpvc: We need to maintain some menu cookies by hand, since enabling one extension may cause others to be enabled, and the menu code only save the cookie for the item that was selected, not the other ones.

extensions/Assistive-Explore.js
```diff
@@ -12,6 +13,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
 
   var Assistive = MathJax.Extension.Assistive = {
     version: '1.0',
+    dependants: [],            // the extensions that depend on this one
```
dpvc: If other extensions rely on this one, and this is disabled, we need to disable the other ones.  So we need to keep a list of the dependents to disable when we are disabled.

extensions/Assistive-Explore.js
```diff
@@ -92,16 +94,46 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
-        'New Math', ['Register', MathJax.Extension.Assistive.Explorer]);
-      if (force) MathJax.Hub.Queue(['Reprocess', MathJax.Hub]);
-    },
-    Disable: function() {
-      if (Assistive.hook) {
-        MathJax.Hub.UnRegister.MessageHook(Assistive.hook);
+    Enable: function(update,menu) {
+      SETTINGS.explorer = true;
+      if (menu) COOKIE.explorer = true;
+      MathJax.Extension.SemanticComplexity.Enable(false,menu);
```
dpvc: Here, we set the `menuSettings` for the explorer to `true`, and if this is a menu action, we set the cookie as well.  Then we signal the Semantic-Complexity extension (on which we rely) to enable itself as well.

extensions/Assistive-Explore.js
```diff
@@ -92,16 +94,46 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+      this.DisableMenus(false);
+      if (!this.hook) {
+        this.hook = MathJax.Hub.Register.MessageHook( 
+          'New Math', ['Register', this.Explorer]);
       }
```
dpvc: Then we enable our submenu entries (disable = false), and if there is not already a global hook, we set one up.

[The reason you couldn't stop the hook was because the `Enable()` was being called more than once, and that means two message hooks were registered, with the second causing you to loose the first value of `this.hook`.  The `Disable()` method then only removed one of the two hooks, so the second continued to be called when the `New Math` message occurred.]

extensions/Assistive-Explore.js
```diff
@@ -92,16 +94,46 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+      if (menu) COOKIE.explorer = false;
+      this.DisableMenus(true);
+      if (this.hook) {
+        MathJax.Hub.UnRegister.MessageHook(this.hook);
+        this.hook = null;
+      }
+      for (var i = this.dependants.length-1; i >= 0; i--) {
+        var dependant = this.dependants[i];
+        if (dependant.Disable) dependant.Disable(false,menu);
+      }
```
dpvc: Here we unset the `menuSettings` value, and unset the cookie if this is a menu action.  We disable the submenu items (disabled = true), and remove the message hook, if any.  Finally, we tell any dependent extensions (none at this point, but maybe there will be in the future) to disable themselves.

extensions/Assistive-Explore.js
```diff
@@ -92,16 +94,46 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+    },
+    DisableMenus: function (state) {
+      if (MathJax.Menu) {
+        var menu = MathJax.Menu.menu.FindId("Accessibility");
+        if (menu) {
+          var items = menu.submenu.items;
+          for (var i = 2, item; item = items[i]; i++) item.disabled = state;
+        }
+      }
+    },
```
dpvc: This function goes through the submenu and disables/enables all the entries (other than the `Active` item at the top).

extensions/Assistive-Explore.js
```diff
@@ -92,16 +94,46 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+    //
+    //  Register a dependant
+    //
+    Dependant: function (extension) {
+      this.dependants.push(extension);
     }
```
dpvc: This allows an extension to register the fact that it depends on Assistive-Explore.

extensions/Assistive-Explore.js
```diff
@@ -238,6 +270,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
     // Registers new Maths and adds a key event if it is enriched.
     //
     Register: function(msg) {
+      if (!Assistive.hook) return;
```
dpvc: The presence of the `hook` indicates that the Assistive-Explore extension is enabled.  So if it is not there, don't register any new math.  (This is just a precaution, since we should now be adding and removing only a single hook, but it would have taken care of the duplicate hook problem in the original code.)

extensions/Assistive-Explore.js
```diff
@@ -314,11 +347,13 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
       math.addEventListener(
           Explorer.focusinEvent,
           function(event) {
+            if (!Assistive.hook) return;
```
dpvc: We make the event handlers check the `hook` (i.e., when the extension is enabled) so we don't have to remove all the hooks in the `Disabled()` method (i.e., we don't have to do a `Reprocess()` call simply to remove the handlers).

extensions/Assistive-Explore.js
```diff
@@ -610,100 +645,104 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
       }
     },
     Startup: function() {
-      SETTINGS['responsive'] = true;
+      var Complexity = MathJax.Extension.SemanticComplexity;
+      if (Complexity) Complexity.Dependant(Assistive);
       Assistive.addDefaults();
-      Assistive.Enable();
+      this.createMenu();
+    },
```
dpvc: The `responsive` settings are now handled in the Semantic-Complexity's `Enable()` method, and we don't want to force it to true in any case, as the user's menu setting may require it to be false.  Similarly, we don't want to force ourselves to be active if the user's menu cookie says otherwise.  (Enabling and disabling are handled below.)

But we do want to tell Semantic-Complexity that we are dependent on it (so if it becomes disabled, it will tell us to disable as well).

extensions/Assistive-Explore.js
```diff
@@ -610,100 +645,104 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+    
+    createMenu: function () {
```
dpvc: The menu code had to be moved to the `Startup()` routine in order to handle the timing right, but now that it is wrapped in the handler for `End Extensions`, that wasn't necessary.  Argh!  I will change that, now that I see it.

extensions/Assistive-Explore.js
```diff
@@ -610,100 +645,104 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+      MathJax.Hub.Register.StartupHook('End Extensions', function () {
+        Assistive[SETTINGS.explorer === false ? "Disable" : "Enable"]();
+        MathJax.Hub.Startup.signal.Post('Explorer Ready');
```
dpvc: We do the Enable/Disable call here so that if the Assistive-Extension has been loaded, it has the chance to set the `SETTINGS` value first.  (If we load Assistive-Explore by itself, that means the page author wants it active by default; but Assistive-Extension turns everything off by default, so we need to wait for Assistive-Explore to run, if it is loaded, before we can tell what the default should be.)  We put off the ready signal until we are actually enabled or disabled.

extensions/Assistive-Explore.js
```diff
@@ -610,100 +645,104 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+        MathJax.Hub.Register.StartupHook('MathMenu Ready', function () {
+          COOKIE = MathJax.Menu.cookie;
+          if (SETTINGS.explorer == null) SETTINGS.explorer = true;
```
dpvc: We save the menu cookie object for later use, and set the `SETTINGS` value if it isn't already.

extensions/Assistive-Explore.js
```diff
@@ -610,100 +645,104 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+          var Switch = function(menu) {
+            Assistive[SETTINGS.explorer ? "Enable" : "Disable"](true,true);
+            MathJax.Menu.saveCookie();
+          };
```
dpvc: The menu action is just to call the `Enable()` or `Disable()` function (informing it to reprocess the page, if needed, and that this action is being done from a menu).  Then we save the menu cookie, in case other extensions were enabled/disabled along the way.

extensions/Assistive-Explore.js
```diff
@@ -610,100 +645,104 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+                      ITEM.RADIO(['chromevox-alternative', 'Alternative'],
+                                 'Assistive-ruleset', speech))
+              );
+          var box = MENU.IndexOfId('Accessibility');
+          if (box !== null) {
+            MENU.items[box] = accessibilityMenu;
+          } else {
+            index = MENU.IndexOfId('CollapsibleMath');
+            MENU.items.splice(index+1,0,accessibilityMenu);
+          }
```
dpvc: Here we find the location of the menu.  It should either replace the existing accessibility menu (created by Assistive-Extension) or should follow the "Collapsible Math" item.

extensions/Assistive-Explore.js
```diff
@@ -610,100 +645,104 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+          Assistive[SETTINGS.explorer ? "Enable" : "Disable"]();
```
dpvc: I think this one can be removed, since the one at 656 should already have done it.

extensions/Assistive-Explore.js
```diff
@@ -610,100 +645,104 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+        },20);  // Between Semantic-Complexity and Semantic-Collapse
+      },20);
```
dpvc: We use priority 20, which is in between the Semantic-Complexity's priority of 15 and Semantic-Collapse's priority of 25.  That means the "Collapsable Math" menu will be there (from Semantic-Complexity), but when Semantic-Collapse adds "Auto Collapse" right after "Collapsible Math", it will move our menu down, and get the right order.

extensions/Assistive-Explore.js
```diff
@@ -610,100 +645,104 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
 });
 
-
-
 MathJax.Ajax.Require("[RespEq]/Semantic-Complexity.js");
 MathJax.Hub.Register.StartupHook('Semantic Complexity Ready', function() {
-  MathJax.Extension.SemanticComplexity.Enable();
   MathJax.Extension.Assistive.Explorer.Startup();
-  MathJax.Hub.Startup.signal.Post('Explorer Ready');
   MathJax.Ajax.loadComplete('[RespEq]/Assistive-Explore.js');
```
dpvc: W don't do the `Enable()` call here (it is done at line 656, and there it is conditional on the state of the menu selections).  We also don't post the ready signal until 657, as discussed there.

extensions/Semantic-Collapse.js
```diff
@@ -3,27 +3,63 @@
...
+    Disable: function (update,menu) {
+      SETTINGS.autocollapse = false;
+      if (menu) COOKIE.autocollapse = false;
+      this.config.disabled = true;
+      for (var i = this.dependants.length-1; i >= 0; i--) {
+        var dependant = this.dependants[i];
+        if (dependant.Disable) dependant.Disable(false,menu);
+      }
+      if (update) HUB.Queue(["Rerender",HUB]);
     },
```
dpvc: These are similar to the ones in Assistive-Explore, except that when we are asked to update, we do that.

extensions/Semantic-Collapse.js
```diff
@@ -3,27 +3,63 @@
...
+    Dependant: function (extension) {
+      this.dependants.push(extension);
+    },
+
     Startup: function () {
       //
+      //  Inform SemanticComplexity that we are a dependant
+      //
+      var Complexity = MathJax.Extension.SemanticComplexity;
+      if (Complexity) Complexity.Dependant(this);
```
dpvc: We register as dependent on Semantic-Complexity.

extensions/Semantic-Collapse.js
```diff
@@ -36,6 +72,35 @@
...
+          var index = MENU.IndexOfId('AutoCollapse');
+          if (index !== null) {
+            MENU.items[index] = menu;
+          } else {
+            index = MENU.IndexOfId('CollapsibleMath');
+            MENU.items.splice(index+1,0,menu);
+          }
+        },25);  // after Assistive-Explore
+      },25);
+    },
```
dpvc: This is similar to Assistive-Explore.  We use `End Extensions` to coordinate among the extensions (using the priorities of the handlers), and handle the configuration from the menu settings or the menu settings from the configuration (depending on whether there was an explicit menu choice).  Then create the menu and place it as described above.

extensions/Semantic-Collapse.js
```diff
@@ -251,51 +317,9 @@
...
-        HUB.Queue(['Reprocess', HUB],
-                  ['resizeHandler', Collapse]);
-      }
-      var item = MathJax.Menu.menu.FindId('Responsiveness', 'Collapse');
-      if (item) {
-        item.disabled = !newResp;
-      }
-      MathJax.Menu.config.settings['collapse'] ?
-        Collapse.Enable() : Collapse.Disable();
-    },
```
dpvc: This routine was never used, even in the original version (which is whey the `resizeHandler` changes that we made last week didn't have any effect).  So we don't need this.

extensions/Semantic-Collapse.js
```diff
@@ -251,51 +317,9 @@
...
-    Remove: function() {
-      Collapse.Disable();
-      MathJax.Extension.SemanticComplexity.Disable();
     }
```
dpvc: This is now handled by `Disbaled()` itself.

extensions/Semantic-Complexity.js
```diff
@@ -8,6 +8,9 @@
 (function (HUB) {
   var MML;
   
+  var SETTINGS = HUB.config.menuSettings;
+  var COOKIE = {};   // replaced when menu is available
+
```
dpvc: The changes here are similar to the ones for Semantic-Collapse above.

extensions/Semantic-MathML.js
```diff
@@ -4,6 +4,7 @@
 MathJax.Extension.SemanticMathML = {
   version: "1.0",
   config: MathJax.Hub.CombineConfig("SemanticMathML",{disabled: false}),
+  dependants: [],     // the extensions that depend on this one
```
dpvc: These changes are similar to the two extensions above.

extensions/Semantic-MathML.js
```diff
@@ -71,6 +89,10 @@ MathJax.Callback.Queue(
   ["Require",MathJax.Ajax,"[MathJax]/jax/input/MathML/config.js"],
   ["Require",MathJax.Ajax,"[MathJax]/jax/input/MathML/jax.js"],
   //
+  //  Load toMathML extension (if it isn't already)(
+  //
+  ["Require",MathJax.Ajax,"[MathJax]/extensions/toMathML.js"],
+  //
```
dpvc: Since this extension relies on toMathML, I force it to load as well.

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-
-})(MathJax.Hub);
+  });
 
+  HUB.Register.StartupHook('End Extensions', function () {
+    HUB.Register.StartupHook('MathMenu Ready', function () {
+      Accessibility.Startup();
+      HUB.Startup.signal.Post('Accessibility Loader Ready');
+    },5);   // run before other extensions' menu hooks even if they are loaded first
+  },5);
```
dpvc: We use the `End Extensions` signal to make sure that the order of menu operations is correct (as discussed in the main comments), since the `MathMenu Ready` signal can occur in essentially random places.  We use the priority (5 in this case, which is less than the default priority of 10) to schedule the Assistive-Extension menus to be made before any of the loaded extensions, which use higher priorities (i.e., later execution).

extensions/Accessibility-Extension.js
```diff
@@ -44,124 +52,104 @@
...
-      var flag = Accessibility.GetOption(this.option);
-      if (flag) {
-        if (this.subMenu) {
-          this.AddSubMenu();
-        } else {
-          MathJax.Ajax.Require(this.module);
-          this.AddMethods();
-          this.AddActiveBox();
-        }
-      }
```
dpvc: We no longer need this `Load()` method (its replacement appears below).

dpvc:
OK, I have made the changes that I mentioned in the line notes.  I think this is ready at this point.  Please check it out and see if you find anything else.

dpvc:
OK, I've pushed some changes that should complete the extension as per our discussion this morning.  It is not as tested as I would like, but I am out of time for this afternoon.  Will do more testing later, but wanted you all to have access to the changes sooner rather than later.

Issue #95, opened 2016-04-28
[a11y extension] location location location


pkra:
I would like to propose to move the dist and build files back into this repo.

Since we're not putting the final extension in the third-party repo, we might as well push them to the CDN from this repo; the other one just adds noise and confusion.

dpvc:
I'm OK with moving things back.

I'm afraid I'm a little out of the loop on where the files are to be stored.  Not in the third-party repository, but on the CDN, somewhere.  What is the URL going to be for that?

I'm assuming the unpacked versions are not going to be available on the CDN, right?  So the `diet` folder represents what will be on the CDN?

pkra:
> What is the URL going to be for that?

I was thinking to still upload this to where the third-party-repo ends up, i.e., `cdn.mathjax.org/mathjax/contrib`.

> I'm assuming the unpacked versions are not going to be available on the CDN, right? So the dist folder represents what will be on the CDN?

That's what I was thinking. But it's not very important to me, 1/10.

pkra:
As per F2F, move things back here, rename this to `mathjax-a11y`

pkra:
I was trying to tackle this (to get into the code and start #97) but the grunt-file seems a bit odd (and misses some resources). 

I've begun writing a fresh one.

@zorkow any suggestions on providing the closure compiler? (Or perhaps publish mathjax-sre as a separate module to npm?)

pkra:
Here's what I came up with. https://gist.github.com/pkra/fe1bd64ba2053665923d07e2747883da

pkra:
Quick smoke test indicates this creates a viable distribution.

pkra:
I've made some minor changes to make repeated builds slightly less of pain.

pkra:
I with #113 merged, I think we can shut down the old repo and rename this one.

Any objections?

dpvc:
OK.  If you are going to do that now, I will wait to make the renaming changes that we discussed in conference today until after that is done.

pkra:
> I will wait to make the renaming changes that we discussed in conference today until after that is done.

I don't think there's a need to wait. GitHub redirects all things. But this can wait until @zorkow is back.

pkra:
I've deleted the old repo (but still have a local copy if we deed one).

I'm confused by myself. Do we call this repo `MathJax-a11y` (stated earlier as result of F2F) or `MathJax-Accessibility` (which I put in the readme)

pkra:
[...]

PR #113, opened 2016-05-30
add build tools

pkra:
See also #95


pkra:
FYI, I've pushed a copy of the dist to https://beta.mathjax.org/mathjax/contrib/a11y and switched http://codepen.io/pkra/pen/yOWrWQ to that.

dpvc:
These don't seem to be the latest versions (they don't have the last two commit that I did last week -- looks like you didn't pull the latest version before making your new branch).  If you add those commits, you will be able to simplify the code pen to only create the path for Contrib, and then load `[Contrib]/a11y/Accessibility-Extension.js` and drop the `RespEq` references entirely.

pkra:
> These don't seem to be the latest versions 

Thanks, I'll update them.

Is the PR ok though?

pkra:
> Thanks, I'll update them.

Update beta.mathjax.org and the codepen example. Looking good.

Issue #114, opened 2016-05-30
Assistive-Explorer.js dummy option


pkra:
Would it be more consistent to name this "none"?

dpvc:
👍

pkra:
Fixed.

Issue #97, opened 2016-04-28
[a11y extension] documentation


pkra:
Here's what I come up with.

* explain the two-in-one extensions or split them up
* document configuration options
* document menu options
* document behavior
  * global aria-label
  * walker 
  * collapse
  * interaction between the various parts, dependencies etc.

dpvc:
I'm not sure what the "two-in-in" extension is.  Is it that the automatic collapsing is also included in this extension as well as the walking?

Would it be better to include only `Semantic-Complexity.js` (which I assume is needed by the walker) rather than `Semantic-Collapse.js`, and allow the page author to include `Semantic-Collapse.js` if desired?

dpvc:
The `Semantic-Collapse` might also be part of the performance problem at start up.  As I recall, it goes through the various collapsings of the equation and records their widths, which means re-rendering the  equations several times.  This probably produces the "New Math" signal, and that may trigger the speech generation for each of those renderings?

pkra:
> I'm not sure what the "two-in-in" extension is. Is it that the automatic collapsing is also included in this extension as well as the walking?

Yes, that's what I meant -- `Assistive-Explore.js` and `Semantic-Collapse.js` wrapped into one.

> Would it be better to include only Semantic-Complexity.js (which I assume is needed by the walker) rather than Semantic-Collapse.js, and allow the page author to include Semantic-Collapse.js if desired?

Assistive-Explore can now use Semantic-Collapse to provide a shorter summary on the top label (perhaps even as default, @zorkow?).

> This probably produces the "New Math" signal, and that may trigger the speech generation for each of those renderings?

@zorkow?

dpvc:
> `Assistive-Explore `can now use `Semantic-Collapse` to provide a shorter summary on the top label 

I don't think it is `Semantic-Collapse` that it is using, but the results of `Semantic-Complexity`, which is why I suggested falling back to that.  @zorkow, are you just using the presence of the complexity mactions (which are created by Semantic-Complexity), or do you use the additional data about the widths of the various possible collapsings (added by Semantic-Collapse)?

dpvc:
[...]

PR #118, opened 2016-06-01
Renaming the extensions

dpvc:
This pull request renames the extension as per our meeting today, and updates the labs and examples to work with the new names paths.  It also adds copyright notices to the extensions.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -22,11 +22,11 @@ module.exports = function(grunt) {
       },
       build: {
         files: {
-          'dist/Accessibility-Extension.js': 'extensions/Accessibility-Extension.js',
-          'dist/Assistive-Explore.js': 'extensions/Assistive-Explore.js',
-          'dist/Semantic-Collapse.js': 'extensions/Semantic-Collapse.js',
-          'dist/Semantic-Complexity.js': 'extensions/Semantic-Complexity.js',
-          'dist/Semantic-MathML.js': 'extensions/Semantic-MathML.js'
+          'dist/accessibility-menu.js': 'extensions/accessibility-menu.js',
+          'dist/explorer.js': 'extensions/explorer.js',
+          'dist/auto-collapse.js': 'extensions/auto-collapse.js',
+          'dist/collapsible.js': 'extensions/collapsible.js',
+          'dist/semantic-enrich.js': 'extensions/semantic-enrich.js'
         }
       }
     },
@@ -51,7 +51,7 @@ module.exports = function(grunt) {
         }
       },
       copy: [
-        'cp speech-rule-engine/lib/sre_mathjax.js dist/',
+        'cp speech-rule-engine/lib/sre_mathjax.js dist/sre.js',
         'cp speech-rule-engine/node_modules/wicked-good-xpath/dist/wgxpath.install.js dist/',
         'cp -R speech-rule-engine/src/mathmaps dist/',
         'cp extensions/*.ogg dist/',
--- a/extensions/accessibility-menu.js
+++ b/extensions/accessibility-menu.js
@@ -22,7 +22,7 @@
       (String(location.protocal).match(/^https?:/) ? "" : "http:") + 
         "//cdn.mathjax.org/mathjax/contrib/a11y");
 
-  var Accessibility = EXTENSIONS.Accessibility = {
+  var Accessibility = EXTENSIONS["accessibility-menu"] = {
     version: '1.0',
     prefix: '', //'Accessibility-',
     default: {},
@@ -120,26 +120,23 @@
   HUB.Register.StartupHook('End Extensions', function () {
     HUB.Register.StartupHook('MathMenu Ready', function () {
       Accessibility.Startup();
-      HUB.Startup.signal.Post('Accessibility Loader Ready');
+      HUB.Startup.signal.Post('Accessibility Menu Ready');
     },5);   // run before other extensions' menu hooks even if they are loaded first
   },5);
   
   Accessibility.Register(
     ModuleLoader(
-      'collapsible', 'Collapsible Math', '[a11y]/Semantic-Complexity.js',
-      'SemanticComplexity'
+      'collapsible', 'Collapsible Math', '[a11y]/collapsible.js', 'collapsible'
     )
   );
   Accessibility.Register(
     ModuleLoader(
-      'autocollapse', 'Auto Collapse', '[a11y]/Semantic-Collapse.js',
-      'SemanticCollapse'
+      'autocollapse', 'Auto Collapse', '[a11y]/auto-collapse.js', 'auto-collapse'
     )
   );
   Accessibility.Register(
     ModuleLoader(
-      'explorer', 'Accessibility', '[a11y]/Assistive-Explore.js',
-      'Assistive', true
+      'explorer', 'Explorer', '[a11y]/explorer.js', 'explorer', true
     )
   );
 
@@ -147,7 +144,7 @@
   
   MathJax.Callback.Queue(
     ["LoadExtensions",Accessibility],
-    ["loadComplete",MathJax.Ajax,"[a11y]/Accessibility-Extension.js"]
+    ["loadComplete",MathJax.Ajax,"[a11y]/accessibility-menu.js"]
   );
 
 })(MathJax.Hub,MathJax.Extension);
--- a/extensions/accessibility-menu.js
+++ b/extensions/accessibility-menu.js
@@ -1,6 +1,27 @@
-//
-// Thin hook to include the accessibility extension.
-//
+/*************************************************************
+ *
+ *  [Contrib]/a11y/accessibility-menu.js
+ *  
+ *  A thin extension to add opt-in menu items for the accessibility
+ *  extensions in the a11y contributed directory.
+ *
+ *  ---------------------------------------------------------------------
+ *  
+ *  Copyright (c) 2016 The MathJax Consortium
+ * 
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
 (function(HUB,EXTENSIONS) {
   var SETTINGS = HUB.config.menuSettings;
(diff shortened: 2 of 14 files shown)


Gruntfile.js
```diff
@@ -22,11 +22,11 @@ module.exports = function(grunt) {
...
       build: {
         files: {
-          'dist/Accessibility-Extension.js': 'extensions/Accessibility-Extension.js',
-          'dist/Assistive-Explore.js': 'extensions/Assistive-Explore.js',
-          'dist/Semantic-Collapse.js': 'extensions/Semantic-Collapse.js',
-          'dist/Semantic-Complexity.js': 'extensions/Semantic-Complexity.js',
-          'dist/Semantic-MathML.js': 'extensions/Semantic-MathML.js'
+          'dist/accessibility-menu.js': 'extensions/accessibility-menu.js',
+          'dist/explorer.js': 'extensions/explorer.js',
+          'dist/auto-collapse.js': 'extensions/auto-collapse.js',
```
pkra: I thought we had decided on `autocollapse` (vs `auto-collapse`) yesterday. I don't mind either way, just wanted to check.

extensions/accessibility-menu.js
```diff
@@ -120,34 +141,31 @@
...
+      'explorer', 'Explorer', '[a11y]/explorer.js', 'explorer', true
     )
   );
 
   Accessibility.AddDefaults();
   
   MathJax.Callback.Queue(
     ["LoadExtensions",Accessibility],
-    ["loadComplete",MathJax.Ajax,"[a11y]/Accessibility-Extension.js"]
+    ["loadComplete",MathJax.Ajax,"[a11y]/accessibility-menu.js"]
```
pkra: When we build a dist for the CDN/Contrib, do we need to change all occurrences to `[Contrib]` (so as to avoid people having to specify two paths when they're using other third party extension)?

pkra: clarification: all occurrences of `[a11y]` to `[Contrib]`

Gruntfile.js
pkra: The documentation also has autocollapse so would need changing.

pkra:
@dpvc do you want to include a fix for #114 (walker: dummy => none)?

extensions/accessibility-menu.js
```diff
@@ -1,6 +1,27 @@
-//
-// Thin hook to include the accessibility extension.
-//
+/*************************************************************
+ *
+ *  [Contrib]/a11y/accessibility-menu.js
```
pkra: WRT to my earlier comment about `[a11y]` in the loadComplete call, I'm now confused :wink:

Gruntfile.js
dpvc: I asked bout changing this in [this post](https://github.com/mathjax/MathJax-RespEq/issues/97#issuecomment-223059076), but didn't get an answer, so I went with my preferred version (hearing no objection).  Yes, I know the the docs would need changing, and will do that if we agree to the name (it sounds like we do).

extensions/accessibility-menu.js
dpvc: No, the extensions all defined `[a11y]` in terms of `[Contrib]` if it isn't already defined (see [lines 41 to 44](https://github.com/mathjax/MathJax-RespEq/blob/0ae8c64ca0b6ab54b109f374f502cd77f5418144/extensions/accessibility-menu.js#L41-L44) for example).  This allows the labs and examples in RespEq to define `[a11y]` to be `../extensions` so that we don't need to have an additional `a11y` subdirectory in `extensions`.  Users will use `[Contrib]/a11y` to access the extensions, but our test files can use `[a11y]` to access a local copy.

dpvc:
> do you want to include a fix for #114 (walker: dummy => none)?

OK.  I also have to turn off AssistiveMML when the explorer is activated.

extensions/accessibility-menu.js
dpvc: In general, `[Contrib]/a11y` and `[a11y]` will point to the same place.  Page authors will use `[Contrib]/a11y` (which is why this is used in the comment), but internally the extensions use `[a11y]` so that we can override that in the labs and examples to use the local copy (which isn't in a directory named `a11y` as would be required for `[Contrib]/a11y`).  That is why `[a11y]` is used in the `loadComplete()` call.  Since `loadComplete()` converts the file name to the full URL anyway, we can use any form that resolves to the correct URL.  In the case where `[a11y]` hasn't be predefined, both `[Contrib]/a11y` and `[a11y]` will point to the same place, and so both will work (the user can load via `[Contrib]/a11y`, and we can `loadComlplete()` via `[a11y]`); when we predefine `[a11y]`, we load via `[a11y]` and the `loadComplete()` does the same.

Gruntfile.js
pkra: > I asked bout changing this in this post, but didn't get an answer, so I went with my preferred version (hearing no objection). 

Sorry, I missed that! No objection from me on dashes.

> Yes, I know the the docs would need changing, and will do that if we agree to the name (it sounds like we do).

:+1:

extensions/accessibility-menu.js
pkra: Ah, thanks for the explanation!

extensions/accessibility-menu.js
pkra: Copy that. Thanks again.

dpvc:
> do you want to include a fix for #114 (walker: dummy => none)

Done.

>  I also have to turn off AssistiveMML when the explorer is activated.

Done.

I think this is ready to go, pending Volker's review.

Gruntfile.js
```diff
@@ -51,7 +51,7 @@ module.exports = function(grunt) {
         }
       },
       copy: [
-        'cp speech-rule-engine/lib/sre_mathjax.js dist/',
+        'cp speech-rule-engine/lib/sre_mathjax.js dist/sre.js',
```
zorkow: sre.js is not a good idea. It might be confused with sre.js that is distributed via npm and that could not be used instead.
We could rename to mathjax-sre.js

dpvc: OK, changed to `mathjax-sre.js` as requested.

dpvc:
I added the patch to SVG, and fixed two other small things.

extensions/explorer.js
```diff
@@ -694,7 +694,7 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
               ITEM.CHECKBOX(['Active', 'Active'], 'explorer', {action: Switch}),
               ITEM.RULE(),
               ITEM.SUBMENU(['Walker', 'Walker'],
-                  ITEM.RADIO(['none', 'No walker'], 'Assistive-walker'),
+                  ITEM.RADIO(['nowalker', 'No walker'], 'Assistive-walker', {value:"none"}),
```
dpvc: the first element of the array is an identifier used by the localization code to find the localized version of the second string in the array, when another language is chosen.  So you need to use unique values that represent the second string.  Since "none" is already associated with the string "None", you need to use something else for "No walker".

The menu code uses the `value` property of the third argument to `ITEM.RADIO()` to set the value used for the menu setting for this radio menu item.  If no value is given, the localization ID is used (since it doesn't depend on the locale).  That is why you don't need to specify the `value` for the other men items.  So unless you want to use `no walker` rather than `none` as the value for not having a walker in the `Assistive-walker` menu setting, we need to specify `value:"none"` here.

pkra:
FYI I ran the Gruntfile without a hitch and have updated the copy on beta.mathjax.org. I've updated the example on codepen and it cleared a quick smoke test.

zorkow:
Re: Gruntfile

- It assumes that a dist directory already exists. In my case I had to create it manually o/w grunt failed. We could simply remove dist in the shell:prepare goal to guarantee a clean built and add a mkdir command. (Or alternatively a mkdir -p.)
- I am not sure I understand the reason for the npm install --only= dev line.
-  Is there a reason why the node_modules are not removed after a build?

O/w it seems to work fine.

pkra:
> It assumes that a dist directory already exists. 

Ah! Originally, I had included dist but then on the F2F I removed it again but didn't remove the dist folder.

> We could simply remove dist in the shell:prepare goal to guarantee a clean built and add a mkdir command. (Or alternatively a mkdir -p.)

Using `mkdir -p` sounds better to me (less destructive).

> I am not sure I understand the reason for the npm install --only= dev line.

We only have `devDependencies` in the package.json. The alternative would be `--production` to install dev. Or we could also just make them regular dependencies.

> Is there a reason why the node_modules are not removed after a build?

If you're on Grunt v1.0+ then grunt is preferrably installed locally. If you then remove the node modules, you would have to install them every time you build. Since `node_modules` only includes the grunt files, it seemed sensible to keep it around (all other dependencies are installed in the SRE folder and thus removed with it).

pkra:
> Ah! Originally, I had included dist but then on the F2F I removed it again but didn't remove the dist folder.

That probably didn't make a lot of sense. Originally, I had added `dist` to git. Then I removed it from git but not from my local file system, so I didn't catch this problem.

zorkow:
Two points: 
* I am always concerned that not everything is fully rebuilt when the old build is not removed.  E.g., on renaming a file an old version might linger on in dist and in the distribution.
* I like to be able to clean my directories from any artefact created by a build. Similar to make clean. 
So we should have at least a grunt goal that allows you to do this manually, but is not in the defaults. 
E.g., `grunt shell:clean_sre` to remove sre in defaults while `grunt shell:clean` will remove everything but is not in defaults.

Also, how do you envision the build process to work: `npm install` and then manually run `grunt`?
If we just want to do an `npm install` then we should have grunt as a prepublish script.

pkra:
> I am always concerned that not everything is fully rebuilt when the old build is not removed. E.g., on renaming a file an old version might linger on in dist and in the distribution.

FWIW, the gruntfile does this the same way your grunt file did on the other repo. But I don't care deeply about this specific behavior.

> I like to be able to clean my directories from any artefact created by a build.  [etc]

Sounds good (though it sounds like the opposite of the previous point).

> Also, how do you envision the build process to work: npm install and then manually run grunt?

Yes.

> If we just want to do an npm install then we should have grunt as a prepublish script.

I obviously didn't see the need but it's not important to me either way.

zorkow:
> > I am always concerned that not everything is fully rebuilt when the old build is not removed. E.g., on renaming a file an old version might linger on in dist and in the distribution.
>
> FWIW, the gruntfile does this the same way your grunt file did on the other repo. But I don't care deeply about this specific behavior.

I've seen that and it's not good there either. I clearly don't adhere to my own principles.

> > I like to be able to clean my directories from any artefact created by a build.  [etc]
>
> Sounds good (though it sounds like the opposite of the previous point).

It's more about when something goes wrong during the build, I don't want to have to clean things up manually. Obviously we don't want to get rid of the dist directory after the build, before we push it somewhere automatically.

> > Also, how do you envision the build process to work: npm install and then manually run grunt?
> 
> Yes.

If that's the intended behaviour, that's fine.

PR #128, opened 2016-06-08
Updates the grunt file with separate cleanup goals.

zorkow:
Adds single goals for each of the cleanup targets.
Adds a clean target that fully purges the repository.


pkra:
LGTM

PR #132, opened 2016-06-14
add documentation

pkra:


dpvc:
Looks good.

One thing I noticed is this is mostly for page *authors*.  Do we need a section on *using* the extensions.  There is some information about the walker, but nothing that says anything about the menus themselves (e.g., a picture showing where the menus are, and indicating what they control), or about ho two collapse/uncollapse equations, or what the collapsed symbols mean.  I know this has appeared in our paper, so I'm sure you are tired of writing that, but do you think something should be included here?

pkra:
Fair point. I'll try to whip something up tomorrow.

pkra:
I've added a user guide with screenshots.

@dpvc I couldn't work out a bookmarklet. Help?

dpvc:
OK, I've identified the problem. When `accessibility-menu` is loaded as part of the page's configuration directly, the startup hooks at [lines 141 to 146](https://github.com/mathjax/MathJax-a11y/blob/master/extensions/accessibility-menu.js#L141-L146) are delayed until after all the extensions load, and so the lines that follow, which define the menus, are run before the hook runs.  When the hook runs, it creates the accessibility menu entries.

On the other hand, when loaded in the bookmarklet, which runs after MathJax has completed its initial startup and typesetting, the "End Extensions" and "MathMenu Ready" signals are already posted, and so lines 141-146 run immediately.  That means the menus haven't been defined yet, since they are created in the lines that follow.  So no menus are inserted.

The solution is to move lines 141-146 to right before line 166.  I will make a pull request for that, but want to do a little more testing first.  But I think that will do it.

Sorry about not getting that timing right in the first place -- it wasn't a case I had considered.

pkra:
@dpvc thanks for looking into it. That means 

```javascript
MathJax.Ajax.config.path["myContrib"] = "https://cdn.mathjax.org/mathjax/contrib";
MathJax.Ajax.Require("[myContrib]/a11y/accessibility-menu.js");
````
will work?

**Comment by pkra** [comment]
I've updated the branch. I think this is ready to merge.

**Comment by dpvc** [comment]
@pkra, yes, though I might suggest

```
MathJax.Ajax.config.path["a11y"] = "https://cdn.mathjax.org/mathjax/contrib/a11y";
MathJax.Ajax.Require("[a11y]/accessibility-menu.js");
```
instead.

pkra:
Fixed.

PR #140, opened 2016-06-15
Move startup hooks to after menu definitions.

dpvc:
Move startup hooks to after menu definitions, so bookmarklet will work (when the signals have already been posted).


pkra:
Works like a charm for the bookmarklet and regularly.

PR #145, opened 2016-07-05
on gh-pages use 3rd-party repository

### LINKED ISSUE #144: fix gh-pages
The extensions fail to load since we don't have include a local sre build in this repo.

Proposal: check the url and set [a11y] to the third party repo if on github.io.

pkra:
Fixes #144

### DIFF
--- a/Semantics-Lab/MathML-linebreaking.html
+++ b/Semantics-Lab/MathML-linebreaking.html
@@ -77,6 +77,9 @@ MathJax.Hub.Config({
   explorer: {speech: false}
 });
 MathJax.Ajax.config.path["a11y"] = "../extensions";
+if(window.location.href.indexOf("mathjax.github.io/MathJax-a11y") > -1) {
+ MathJax.Ajax.config.path["SRE"] = "https://cdn.mathjax.org/mathjax/contrib/a11y";
+}
 
 (function() {
   window.addEventListener("resize", resizeThrottler, false);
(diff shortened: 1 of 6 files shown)


pkra:
Yikes. Sorry for auto-correcting of whitespace.

dpvc:
How about setting `[SRE]` rather than `[a11y]` since it is only SRE that is not in the github directory?

pkra:
> How about setting [SRE] rather than [a11y] since it is only SRE that is not in the github directory?

My rationale was that gh-pages should be a bit more stable and will likely only be updated upon releases.

But I'm happy to change it -- it's a 3/10 for me.

pkra:
@dpvc *bump* would this be more than a 3/10 for you?

dpvc:
No, `main.js` is fine.

pkra:
I think you're thinking of the mathjax-node PR :-) This is about mathjax-a11y.

dpvc:
Sorry, yes I was.  I think it is better to use `[SRE]` since we have in the past used the `github.io` pages to demo development features, which may not yet be in the release version on the CDN.  I know that we aren't doing much development here now, but if you set `[a11y]`, then really all you are getting is the demo page, not the actual extensions from `github.io`, so why pull the whole repo into it?

I'm probably 6/10 for this.

pkra:
Thanks! I'll update the PR in the morning.

pkra:
I've amended the commit. PTAL.

PR #153, opened 2016-09-06
Move accessibility menu items to a submenu

dpvc:
This PR makes a change that I'm proposing to be included in 2.7 that moves the accessibility menu items to an accessibility submenu rather than putting them in the main menu.  This makes the accessibility settings work like the Math Settings, and doesn't clutter the main menu so much.  This would include a change to the MathMenu.js file to create the initial Accessibility submenu, and would move the "Assistive MathML" and "Include in Tab Order" menu items from the Renderer submenu to the Accessibility menu, so the menu is always populated with something, even if the accessibility-menu extension isn't loaded.  The extension then adds the three new items to that menu when it *is* loaded.

Here is how it looks:

menu

It still works with v2.6; the extension will create the Accessibility menu item itself and move the two existing items to it automatically.  (In 2.6, if the accessibility-menu extension isn't loaded, but the individual extensions are, then they will add their items to the main menu, as they do now.)

If you don't like the idea of the accessibility menu always being there, then we could leave the AccessibleMML and InTabOrder items in their current places, and the extension would make the new submenu (and move the two items) automatically.  But I think it may be better to have the accessible menu all the time, as I never really thought those two items fit as well in the renderer menu anyway.

In any case, I think it is worth moving the new accessibility items to a submenu.  What do you think?

### DIFF
--- a/extensions/explorer.js
+++ b/extensions/explorer.js
@@ -752,12 +752,18 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
                              'Assistive-ruleset', speech)
               )
           );
-      var box = MENU.IndexOfId('Explorer');
-      if (box !== null) {
-        MENU.items[box] = explorerMenu;
+      var submenu = (MENU.FindId('Accessibility')||{}).submenu, index;
+      if (submenu) {
+        index = submenu.IndexOfId('Explorer');
+        if (index !== null) {
+          submenu.items[index] = explorerMenu;
+        } else {
+          index = submenu.IndexOfId('CollapsibleMath');
+          submenu.items.splice(index+1,0,explorerMenu);
+        }
       } else {
-        var index = MENU.IndexOfId('CollapsibleMath');
-        MENU.items.splice(index + 1, 0, explorerMenu);
+        index = MENU.IndexOfId('CollapsibleMath');
+        MENU.items.splice(index+1,0,explorerMenu);
       }
       if (!SETTINGS.explorer) Assistive.DisableMenus(true);
     },20);  // Between collapsible and auto-collapse extensions
--- a/extensions/explorer.js
+++ b/extensions/explorer.js
@@ -752,18 +752,18 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
                              'Assistive-ruleset', speech)
               )
           );
-      var submenu = (MENU.FindId('Accessibility')||{}).submenu, index;
+      var submenu = (MENU.FindId('Accessibility') || {}).submenu, index;
       if (submenu) {
         index = submenu.IndexOfId('Explorer');
         if (index !== null) {
           submenu.items[index] = explorerMenu;
         } else {
           index = submenu.IndexOfId('CollapsibleMath');
-          submenu.items.splice(index+1,0,explorerMenu);
+          submenu.items.splice(index + 1, 0, explorerMenu);
         }
       } else {
         index = MENU.IndexOfId('CollapsibleMath');
-        MENU.items.splice(index+1,0,explorerMenu);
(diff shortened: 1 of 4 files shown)


zorkow:
I suppose it makes sense. When we included the "include in taborder" we did not really know where to put it and the current place was a stop-gap. I would assume so was the positioning of the assistivemml command. So having a single, topical submenu that collates all the assistive and alternative access options makes sense to me.

I can do a proper code review tomorrow and test the reload functionality.

dpvc:
Thanks!

zorkow:
I notice that on switching on Speech Output, the subtitles command remains greyed out.  Only on page reload subtitles is selectable.
This behaviour seems to exist already before the PR. Is it worthwhile fixing it before 2.7?

extensions/explorer.js
```diff
@@ -752,12 +752,18 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
                              'Assistive-ruleset', speech)
               )
           );
-      var box = MENU.IndexOfId('Explorer');
-      if (box !== null) {
-        MENU.items[box] = explorerMenu;
+      var submenu = (MENU.FindId('Accessibility')||{}).submenu, index;
```
zorkow: Whitespace around ||

(This being my part of the code, I have to insist :-)

extensions/explorer.js
```diff
@@ -752,12 +752,18 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+      if (submenu) {
+        index = submenu.IndexOfId('Explorer');
+        if (index !== null) {
+          submenu.items[index] = explorerMenu;
+        } else {
+          index = submenu.IndexOfId('CollapsibleMath');
+          submenu.items.splice(index+1,0,explorerMenu);
```
zorkow: More whitespace...

extensions/explorer.js
```diff
@@ -752,12 +752,18 @@ MathJax.Hub.Register.StartupHook('Sre Ready', function() {
...
+        }
       } else {
-        var index = MENU.IndexOfId('CollapsibleMath');
-        MENU.items.splice(index + 1, 0, explorerMenu);
+        index = MENU.IndexOfId('CollapsibleMath');
+        MENU.items.splice(index+1,0,explorerMenu);
```
zorkow: ditto.

zorkow:
Thanks for the whitespace!I feel it belongs after punctuation;otherwise,it looks very odd,does it not?

+1 to the change in MathJax to move accessibility menu consistently in the same spot. Otherwise, it might give an inconsistent UX.

What are we going to do with the subtitles entry problem?

pkra:
> +1 to the change in MathJax to move accessibility menu consistently in the same spot. Otherwise, it might give an inconsistent UX.

+1

dpvc:
I have made a pull request to make the menu change in MathJax.  This pull request should be merged before 2.7 is released (the old version of accessibility-menu will work either way, but the old version will not use the new Accessibility submenu available in the MathJax patch).

pkra:
Do we need to make a new release of the a11y extensions?

dpvc:
Yes, please.

pkra:
Ok. Could somebody suggest an tag number (1.0.1 / 1.1.0 / 2.0.0) and a short release note with changes? I've lost track of what was updated.

dpvc:
I guess it should be 1.1.0.  I'll have to change that in the files.

The change is that the three menu items now appear in a separate submenu (the `Accessibility` menu), along with the `Assistive MathML` and `In Tab Order` items.  That menu is new in v2.7 (and moves the two other items from the `Math Settings->Math Renderer` submenu; in v2.6, the accessibility extension creates the menu and moves those items.

dpvc:
I've bumped the version numbers.

Issue #154, opened 2016-10-05
Issues with collapsing during exploration  (math.SE)


pkra:
I was just testing math.stackexchange (e.g [this](http://math.stackexchange.com/questions/1954548/how-do-i-perform-a-first-order-taylor-expansion-on-fracfx-1dx-1fx-2d)) and noticed the following bugs:
- collapsing during walking ends the walker 
- collapsed equations did not get a summary

pkra:
For the record, I saw this on Chrome (Linux) and Firefox (Win7).

zorkow:
1. When you say collapsing, do you mean using keyboard or mouse clicks? In the latter case the walker will indeed stop (can be restarted in the previous position, however, if this position still exists). The problem is that the mouse click can occur anywhere, i.e., on the highlighted element but also above, below or outside. One would therefore have to recompute the walker entirely and it's not even clear how far one would have to walk. We discussed that at the time and decided it is not worse the hassle.

2. Do you mean the aria-label?  I did not see that. (I've been testing with Chrome and Firefox in Linux).
Can you point me to a concrete example where that happens?
Note also that we made sure that summary works well with the MathSpeak rules but the old ChromeVox rules do not necessarily always play together well with the summary rules.

pkra:
> When you say collapsing, do you mean using keyboard or mouse clicks?

Keyboard.

> Do you mean the aria-label?

No. Once I re-entered the walker, it did continue at the collapsed expression but that expression did not get a summary corresponding to the collapse but voiced the entire subexpression.

pkra:
> Can you point me to a concrete example where that happens?

I saw this on math.stackexchange.com. Eg., http://math.stackexchange.com/questions/1972241/lim-x-rightarrow-0-frac1-cos-a-1x-cdot-cos-a-2x-cdot-cos-a-3x-cdo, the limit in the question.

zorkow:
I've played with a number of Browser/OS combinations on SauceLabs and I can not reproduce the problem. Please have a look for yourself. Are we sure that all files are fully purged, caches have been cleared, etc.?
Also, what was the particular error message that you saw?

zorkow:
Well I can also not reproduce the SVG collapse failure locally. However, what I can see is that the mactions in SE don't have any class attribute.  E.g.,
document.getElementsByClassName('mjx-svg-maction');
does not yield any results.
@dpvc is there a setting we are missing? The highlighter relies on the class name being there.
[...]

PR #160, opened 2016-10-25
Fixes SVG collapsing problem which is part of issue #154.

zorkow:
As discussed this morning.

### DIFF
--- a/extensions/auto-collapse.js
+++ b/extensions/auto-collapse.js
@@ -410,7 +410,6 @@ MathJax.Hub.Register.StartupHook("HTML-CSS Jax Ready",function () {
 //  Specific implementations for SVG output.
 //
 MathJax.Hub.Register.StartupHook("SVG Jax Ready",function () {
-  MathJax.Hub.Config({SVG: {addMMLclasses: true}});
   MathJax.OutputJax.SVG.Augment({
     getMetrics: function (jax) {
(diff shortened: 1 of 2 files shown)


zorkow:
Ping.

dpvc:
I had made a comment earlier, but apparently hadn't submitted it properly (I need to learn more about this new "review" feature of github).  I think I have submitted it correctly now.  Sorry for the error (the caused the delay).

extensions/auto-collapse.js
```diff
@@ -410,7 +410,6 @@ MathJax.Hub.Register.StartupHook("HTML-CSS Jax Ready",function () {
 //  Specific implementations for SVG output.
 //
 MathJax.Hub.Register.StartupHook("SVG Jax Ready",function () {
-  MathJax.Hub.Config({SVG: {addMMLclasses: true}});
```
zorkow: Had not thought of that. But does the auto-collapse really need the MML classes? Given that we can have collapse without auto-collapse, I would assume it is not strictly necessary. 
It shouldn't do any harm having it, though.

dpvc: Now that you mention it, I'm not sure.  It has been a while since I worked on the code, and I don't remember if there was a need for the in the auto collapsing or whether it was just left over from earlier versions that set it for the walker to work.  I'll have to check again...

dpvc: OK, it looks like it is OK to remove this line after all.

Issue #163, opened 2016-11-26
Unicode symbol for embellished relations


zorkow:
Relation symbols made up of more than one symbol get a mathml-like element as collapse symbol:
a =_2 b =_2 c =_2 d
collapses to
◂mo(=)mn(2)⋯▸

On the other hand embellished operators work. E.g.,
a +_2 b +_2 c +_2 d
collapses to
◂+▸

dpvc:
I have made a pull request that should fix this.

pkra:
This was fixed via #169

PR #169, opened 2016-12-01
Make sure FindChildText() uses the core mo

dpvc:
Make sure FindChildText() uses the core mo, if there is one.  This makes sure that the collapsed marker uses the text of the core element rather than tries to use the nested MathML for embellished operators.

Resolves issue #163.


zorkow:
lgtm.

Issue #166, opened 2016-11-29
Subtitles are grayed out when first activating explorer


pkra:
Regression from https://github.com/mathjax/MathJax-a11y/pull/153#issuecomment-245326299.

As per F2F, https://github.com/mathjax/MathJax-a11y/blob/master/extensions/explorer.js#L653 needs an update.

dpvc:
=> Merged.

PR #170, opened 2016-12-02
Fix disabling of explorer menus.  #166

dpvc:
Fix enabling/disabling of subtitles and generation when speech output is toggled, and fix enabling/disabling of menus when activated/deactivated.  Resolves issue #166.


zorkow:
LGTM.

Issue #173, opened 2017-03-02
Enabling auto-collapse without enabling collapsible-math results in undefined variable usage


mitya57:
Steps to reproduce (with Chromium):

* Open any page with math (i.e. one on mathjax.org main page);
* Enable Chromium dev tools;
* Select Accessibility → Auto Collapse.

At this point Chromium warns about undefined variable usage. This happens in [auto-collapse.js:51](https://github.com/mathjax/MathJax-a11y/blob/master/extensions/auto-collapse.js#L51):
```js
MathJax.Extension.collapsible.Enable(false,menu);
```
here `MathJax.Extension.collapsible` is undefined.

pkra:
Thanks for the report!

dpvc:
The auto-collapse extension does load the collapsible extension, but it looks like the `Enable()` function is being called before the other one has finished loading.  I'll have to look into it.  The interconnections among the various components is pretty complicated, since they can be loaded independently or together, and need to load automatically on menu selections, etc.  But something is clearly amiss.

dpvc:
The `issue173` branch includes a patch for this.

dpvc:
==> Merged.

Issue #175, opened 2017-03-14
Allow explorer without collapse


pkra:
Raised on the [user group](https://groups.google.com/forum/#!topic/mathjax-users/uS-O0aW7_O4).

It might be useful to allow the explorer to work without the collapsing extension.

jfeferman:
Collapsible is an excellent tool for complex equations but may be overkill on content populated with simple equations (e.g. no reason to collapse "y = x + 1" while you may still want to use explorer for accessibility purposes). In some cases user experience could be enhanced by disabling the collapse feature while explorer is active.

dpvc:
Just to be clear, collapsible does not make *everything* collapsible, only those subexpressions that are complicated enough.  So `y = x + 1` doesn't actually become collapsible.  Also, collapsible doesn't actually cause anything to *be* collapsed; it just makes it *possible* for them to be collapsed, if the user clicks on them.  So no expressions will be collapsed initially, and they only collapse if the reader specifically collapses them.  Finally, note that the explorer has keyboard shortcuts for collapsing and uncollapsing sub-expressions, and so even if it is collapsed and you want to explore it, you can.

Can you be more specific about how the user experience is enhanced by disabling the collapse feature during exploration?  It is not clear what you are concerned about specifically.

jfeferman:
Thank you for your reply. Absolutely, collapsible works as you describe and is triggered by a user initiated action, either via mouse or keyboard. My example equation and simple expressions do not, as you point out, trigger collapsible but anything as simple as a quadratic equation does. Many equations in my project do trigger collapsible. 

My concern is that equation collapsing provides an extraneous feature to the core content and page interaction in my project. This could be viewed as a distraction and not add to the user experience, at the level of math we employ. Providing an option to disable this feature could be useful to developers, depending on the intended use and complexity of math.

drlippman:
@dpvc, you said above that "no expression will be collapsed initially", but my experience was that while an equation like `f(x)=x^2` was not visually collapsed, the screenreader did read it as "collapsed equality."  Perhaps that's an unrelated issue, though.

zorkow:
You are right, David. What is presented to a screen-reader in browse mode, exposed through an aria-label, is the most concise reading of a formula, i.e., the most collapsed version, regardless of whether a formula is visually collapsed or not.
The reason for this decision was that speaking every formula in its full glorious details adds considerably to the time it takes to screen read a document, not to mention the added cognitive load of trying to understand the content.

[...]

PR #184, opened 2017-04-13
Do proper synchronization between collapse and auto-collapse. #173

dpvc:
Properly synchronize enabling of collapse extension when auto-collapse is loaded via the menu.

Resolves issue #173.


dpvc:
> For some reason switching on explorer always switches on collapsible. Should that be the case?

Yes.  Explorer loads collapsible (as I recall, it uses the maction elements that collapsible), and so it is enabled.  There is an open issue (#175) asking for explorer without collapse.

Issue #180, opened 2017-04-10
remove references to cdn.mathjax.org


pkra:

dpvc:
==> Merged.

PR #185, opened 2017-04-13
Update for change in cdn

dpvc:
Update to use copy in extensions if there is no a11y path specified.

After this is merged, we need to run grunt, and move the resulting files into the MathJax repo in the `extensions/a11y` directory


dpvc:
Resolves #180

Issue #202, opened 2017-06-16
Use of 'default' as property name causes problems in old IE


dpvc:
The us of `default` as a property in the accessibility-menu extension (which is now included in all the default configurations) causes IE8 to fail when loading the configuration, and that prevents MathJax from running in IE8.  That is true whether any of the accessibility extensions are active or not (since it is in the menu extension itself).

dpvc:
==> Merged.

PR #203, opened 2017-06-16
Don't use default as a property name (for IE8 and below).

dpvc:
This PR changes the occurrences of `default` as a property name to another name, since "default" used to be a reserved word, and it causes IE8 and below to fail when loading the accessibility-menu extension.

Resolves issue #202.


zorkow:
lgtm.

Issue #208, opened 2017-06-19
New SRE version 


zorkow:
Integrate latest SRE (v2.0.0-beta.3) that uses js instead of json file extensions for speech mappings,  to avoid cdnjs problems. 

@dpvc Do we need a new MathJax point release for that?

pkra:
> Do we need a new MathJax point release for that?

Either way, we need a new MathJax-a11y release.

pkra:
E.g., https://github.com/mathjax/MathJax-a11y/blob/master/Gruntfile.js#L13 needs to be changed.

pkra:
This was fixed  via #209

PR #209, opened 2017-06-20
Issue208

zorkow:
Rebuilds distribution to integrate new SRE as discussed in issue #208.


Gruntfile.js
```diff
@@ -10,8 +10,8 @@ module.exports = function(grunt) {
       'speech-rule-engine': {
         options: {
           repository: 'https://github.com/zorkow/speech-rule-engine.git',
-          branch: 'v2.0.0-beta.2',
-          cwd: "."
+          branch: 'v2.0.0-beta.3',
```
pkra: I think this should wait until v2 is out of beta.

zorkow: A cheap shot.... 
But seriously here is my point:
I hated renaming JSON files to .js files. If it does not solve the problem I would revert that change before the 2.0 release. So I would like to test this first.
Given that there will be no more changes between beta.3 and 2.0 if it works, all we would need to do is update this line of code, without having to do another MathJax point release.

Would we really need another

pkra: > Given that there will be no more changes between beta.3 and 2.0 if it works

Famous last words :wink: 

Ok with me then.

zorkow: I've already written the release notes https://github.com/zorkow/speech-rule-engine/wiki/Release-Notes-v2.0
There might still be an update to the README, but that's it!

dpvc: I'm OK with it as well.

dpvc:
I think we can move on this.

PR #237, opened 2020-05-29
Upgrade a11y extension to work with latest SRE

zorkow:
Adapts the code to work with the latest version of SRE. __Note, that this is not backward compatible and will only work with SRE >=3.0__

In detail it: 
* Converts to the lazy computation of walkers (unfortunately SRE does not use getters o/w this would be compatible).
* Maps old collapse key `Return` to new key `-`.
* Enables the keyboard cycling of rule sets, and thus allows the use of Clearspeak in v2. Now we do not have facilities to update the menu as well, thus menu and speech rule set settings can get out of sync. This is the final commit, we can leave it off if you don't think that is a good idea.

### DIFF
--- a/extensions/accessibility-menu.js
+++ b/extensions/accessibility-menu.js
@@ -41,7 +41,7 @@
   if (!PATH.a11y) PATH.a11y = HUB.config.root + "/extensions/a11y";
 
   var Accessibility = EXTENSIONS["accessibility-menu"] = {
-    version: '1.5.0',
+    version: '1.6.0',
     prefix: '', //'Accessibility-',
     defaults: {},
     modules: [],
--- a/extensions/auto-collapse.js
+++ b/extensions/auto-collapse.js
@@ -33,7 +33,7 @@
   if (!PATH.a11y) PATH.a11y = HUB.config.root + "/extensions/a11y";
 
   var Collapse = MathJax.Extension["auto-collapse"] = {
-    version: "1.5.0",
+    version: "1.6.0",
     config: HUB.CombineConfig("auto-collapse",{
       disabled: false
     }),
--- a/extensions/collapsible.js
+++ b/extensions/collapsible.js
@@ -41,7 +41,7 @@
   if (!PATH.a11y) PATH.a11y = HUB.config.root + "/extensions/a11y";
 
   var Collapsible = MathJax.Extension.collapsible = {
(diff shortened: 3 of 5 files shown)


dpvc:
This works for me, and I was able to build the new version without issue.  Thanks!

Two things.  First, in v3, hitting return when the equation is focused enters the explorer.  Here, I must use shift-space (which also works in v3).  Should return work in v2?  (Or is v3 in error to allow it?)

Second, what changes to the menu would be needed for the Clearspeak rules?  There are mechanisms for changing the menu on the fly, so we might be able to do what is needed.  Is it worth looking into that?

zorkow:
> Two things. First, in v3, hitting return when the equation is focused enters the explorer. Here, I must use shift-space (which also works in v3). Should return work in v2? (Or is v3 in error to allow it?)
> 

We made the deliberate decision to move to Return in v3, keeping Shift-space for backward compatibility. I did not enable return in v2, but could do if you like.

> Second, what changes to the menu would be needed for the Clearspeak rules? There are mechanisms for changing the menu on the fly, so we might be able to do what is needed. Is it worth looking into that?

Enabling the smart preferences should probably not be too difficult. The relevant method is:
```JavaScript
sre.ClearspeakPreferences.smartPreferences(item, locale)
```
Here item is a v3 MathItem. We only need that to get the current speech explorer, though. In v2 we only ever had one explorer (i.e., not multiple per item). So we could have a method that just takes explorer + locale. It returns a set of menu items in JSON format; we would have to rewrite them.

zorkow:
> > Two things. First, in v3, hitting return when the equation is focused enters the explorer. Here, I must use shift-space (which also works in v3). Should return work in v2? (Or is v3 in error to allow it?)
> 
> We made the deliberate decision to move to Return in v3, keeping Shift-space for backward compatibility. I did not enable return in v2, but could do if you like.

I've enabled this now. So `return` and `shift-space` have the same effect.

> 
> > Second, what changes to the menu would be needed for the Clearspeak rules? There are mechanisms for changing the menu on the fly, so we might be able to do what is needed. Is it worth looking into that?
> 

I've added a Clearspeak radio button, so it can be switched to in the menu. The menu is now also updated on keyboard switching rule sets. However, changes to the clearspeak preferences can only be done via keyboard, so there is no submenu.

